PR #2, opened 2020-11-16
Initial stubbed API Gateway

danielgrant:
This PR delivers a stubbed / mocked API Gateway defined in Terraform. It uses stubbed Lambda Functions to return dummy data, and contains JSON schemas to define the inputs / outputs from the API Gatway endpoints, as well a custom token authorizer.


.pre-commit-config.yaml
```diff
@@ -0,0 +1,18 @@
+- repo: git://github.com/antonbabenko/pre-commit-terraform
```
kenoir, line 1: This is nice - not seen this before.

kenoir (approved):
Looks good to me 👍

infra/lambda-stub.tf
```diff
@@ -0,0 +1,203 @@
...
+  function_name = aws_lambda_function.stubs_authorizer.function_name
+  principal     = "apigateway.amazonaws.com"
+  source_arn    = "${aws_api_gateway_rest_api.identity.execution_arn}/*/*"
+}
+
+# stubs/empty-200.js
+
+resource "aws_lambda_function" "stubs_empty-200" {
+  function_name    = "identity-stubs-empty-200-${terraform.workspace}"
+  filename         = "${path.module}/../lambda/src/stubs/empty-200.js.zip"
```
garyttierney, line 39: The pattern I've seen for deploying Lambdas across existing Wellcome infrastructure is to store the Lambda zip in an S3 bucket and configure the function to point to that instead of a local file (see https://github.com/wellcomecollection/goobi-infrastructure/blob/588dad11bf24693aea02c0fc786c01b09dfda0de/infrastructure/staging/lambdas.tf, https://github.com/wellcomecollection/search-logger/blob/a3238dd3195f9424c8b75e73052ba93a211621af/terraform/lambda.tf).

I think this is the approach we should be going for. It keeps the build process decoupled from the deployment process in the same way that producing the Docker image as a build artifact and consuming it later in the deployment would be. Right now you will always need to re-build the Lambdas with the currently deployed version of the source code to avoid clobbering the previously deployed Lambda.

Not a problem in itself, but can be a bit of a nuisance. As a bonus, rollbacks also become possible without rebuilding the commit you want to rollback to if we key those zip files with a version number (e.g. users-list-v0.1.0.js.zip).

This might be better placed for when we have the TypeScript project and build in place (see https://github.com/wellcomecollection/search-logger/blob/d191e397c80814d909cbefd11971081402485980/lambda/package.json#L7 for an example of that), so I could pick it up as part of my work, but interested on your thoughts on this.

danielgrant: I don't disagree and happy for this to be picked up when we start delivering the Lambda's properly - for the purposes of these stubs, we can stick with the noddy approach in this PR.

kenoir: Nice spot @garyttierney.

infra/scoped/api-documentation.tf
```diff
@@ -0,0 +1,585 @@
...
+    aws_api_gateway_documentation_part.users_userid_lock,
+    aws_api_gateway_documentation_part.users_userid_lock_put,
+    aws_api_gateway_documentation_part.users_userid_unlock,
+    aws_api_gateway_documentation_part.users_userid_unlock_put
+  ]
+}
+
+# API
+
+resource "aws_api_gateway_documentation_part" "api" {
```
garyttierney, line 31: Since you've already done all the work in creating the JSON schemas it'd be nice to have this stored as an OpenAPI file instead of Terraform resources so you can work on it locally. AWS does [support](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-documenting-api-quick-start-import-export.html) that, but not via Terraform it seems. Maybe one for later.

garyttierney (approved):
LGTM. There is a lot of duplication between `api_documentation` and `api_method` resources, but I don't think it's worth trying to reduce it at this point. The configuration seems fairly straightforward.

kenoir:
Nitpick: We usually call what you've named`backend.tf`: `terraform.tf`.

kenoir:
The way we've managed "scoped" things elsewhere is to have "critical" and "stack" modules, then instantiate them per environment, see: https://github.com/wellcomecollection/storage-service/tree/master/terraform.

PR #5, opened 2020-11-16
Implementing custom hostname over v1 of the API Gateway

danielgrant:
Also fixes a typo in the delegated DNS zone.


infra/scoped/api-stage.tf
```diff
@@ -1,10 +1,28 @@
...
+  stage_name    = local.identity_v1
+}
+
+resource "aws_api_gateway_domain_name" "identity_v1" {
+  certificate_arn = aws_acm_certificate_validation.identity_api_v1.certificate_arn
+  domain_name     = local.identity_v1_hostname
+
+  tags = merge(
+    local.common_tags,
+    map(
```
garyttierney, line 18: You can use map literals in Terraform now BTW. e.g, `{name = local.identity_v1_hostname}`

garyttierney (approved):
LGTM

infra/scoped/api-stage.tf
danielgrant: Good shout, effecting that change in https://github.com/wellcomecollection/identity/pull/6.

PR #6, opened 2020-11-16
Initial Auth0 configuration to support identity services

danielgrant:


garyttierney (approved):
Is auth0 a core provider? i.e. you don't need to do any local setup to get it working? LGTM.

infra/scoped/provider.tf
```diff
@@ -16,3 +34,7 @@ provider "template" {
 provider "archive" {
   version = "~> 2.0.0" # Any non-beta version >= 2.0.0 and <2.1.0, i.e. 2.0.X
 }
+
+provider "auth0" {
```
kenoir, line 38: You might want to split out Auth0 things into their own stack, every time we run this it'll require Auth0 credentials, we'd expect to change that less often then other infra config.

danielgrant: I'm not clear on how splitting this out into its own stack helps with the availability of credentials when `apply`'ing the Terraform? Those credentials are defined as applications within the Auth0 tenant that have access to the API and are appropriately scoped.

In my case, I have a local script that initialises the environment variables `AUTH0_DOMAIN`, `AUTH0_CLIENT_ID` and `AUTH0_CLIENT_SECRET` environment variables. This is easier and requires less clicks of the keyboard than e.g. authenticating against AWS on the CLI...

kenoir: We tend to configure the AWS provider like this: https://github.com/wellcomecollection/storage-service/blob/master/terraform/stack_prod/provider.tf#L2

The mechanism we use to [authenticate with AWS](https://github.com/wellcomecollection/platform-infrastructure/tree/master/accounts) requires authenticating only about once per day for all CLI AWS tasks and we just assume role depending on situation.

Splitting the Auth0 stuff out would mean that we wouldn't need to worry about having those env vars set up when we run the stack - either that or perhaps provide the script to setup the env vars, or add a comment indicating that's required.

PR #7, opened 2020-11-17
Configured a custom domain for Auth0

danielgrant:


infra/scoped/output.tf
```diff
@@ -1,15 +1,22 @@
...
-  value = auth0_client.dummy_test.client_id
+output "auth0_endpoint" {
+  value = "https://${auth0_custom_domain.identity.domain}/"
 }
 
-output "auth0_client-dummy-client_secret" {
-  value = auth0_client.dummy_test.client_secret
+# Auth0 clients
+
+output "auth0_client_dummy-test" {
```
garyttierney, line 15: The mixing of `_` and `-` in identifiers seem a bit odd. Any specific reason for it?

garyttierney (approved):

infra/scoped/output.tf
danielgrant: Yeah, I try to use `_` in lieu of a space, and a hyphen as they would normally we used or in place of a `/` etc.

So `auth0_client_dummy-test` could be interpreted as `Auth0 Client dummy-test`.

PR #12, opened 2020-11-24
Implemented basic Lambda stubs

danielgrant:


packages/apps/api/src/app.ts
```diff
@@ -1,17 +1,94 @@
...
+  console.log('Processing [GET /users/:user_id]...');
+  res.status(200).json(user_one);
+});
+
+app.put('/users/:user_id', function (req: Request, res: Response) {
+  console.log('Processing [PUT /users/:user_id]...');
+  res.status(200).json(user_one);
+});
 
+app.delete('/users/:user_id', function (req: Request, res: Response) {
```
garyttierney, line 34: Ah, I see why you need to call `json` here. A response needs to be ended: https://expressjs.com/en/api.html#res.end. Sending data implicitly does it for you.

garyttierney (changes requested):

garyttierney (approved):

PR #14, opened 2020-11-24
Implements integration with Sierra API to support Sierra credential validation

danielgrant:


packages/shared/sierra-client/src/index.ts
```diff
@@ -77,57 +105,89 @@ export default class SierraClient {
...
 
-  private extractVarField(varFields: VarField[], fieldTag: string) {
+  private extractVarFieldContent(varFields: VarField[], fieldTag: string) {
     const found = varFields.find(varField => varField.fieldTag === fieldTag);
     return found ? found.content : '';
   }
 
-  private getPatronName(varField: string): PatronName {
-    if (!varField.trim()) {
+  private getPatronName(varFields: VarField[]) {
```
garyttierney, line 134: I think this should go out into it's own module so we can unit test deserialization of Sierra records.

packages/shared/sierra-client/src/index.ts
```diff
@@ -77,57 +105,84 @@ export default class SierraClient {
...
     }).then(response => {
-      return axios.create({
+      const instance = axios.create({
         baseURL: this.apiRoot,
         headers: {
           Authorization: 'Bearer ' + response.data.access_token
-        },
+        }
       });
+      instance.interceptors.response.use(response => response, error => this.unhandledError(error));
```
garyttierney, line 114: Errors are turned into whatever `this.unhandledError(...)` is returning here, but the error is never caught by checking the promise for errors.

garyttierney (changes requested):

packages/shared/sierra-client/src/index.ts
danielgrant: That's not quite right. This construct here is defining the default handler for anything which Axios construes as an erroneous response, i.e. request errors such as timeouts, or any response with an HTTP status code that doesn't match the `validateStatus` function. When such errors occur, the caller will receive our custom `ErrorResponse` type, with a `status` of `SierraStatus.UnknownError` and a descriptive error message in the `message` field.

There's no error / exceptions being thrown around here, so `try / catch` or `Promise`'s `.catch(...)` doesn't apply.

danielgrant: With that being said, there is a problem here - `this.unhandledError(error)` will receive an `AxiosError`, not an `AxiosResponse`. Will fix that.

danielgrant: I've updated the status code / error handling approach to resolve this.

packages/shared/sierra-client/src/index.ts
danielgrant: I've refactored the Patron models out into `patron.ts`.

garyttierney (approved):

PR #15, opened 2020-11-25
Added SES email configuration for Auth0

danielgrant:


infra/static/output.tf
```diff
@@ -3,3 +3,26 @@
...
   value = aws_route53_zone.identity.name_servers
 }
+
+# SES
+
+output "wellcomecollection_org_ses_vertification_token" {
+  value = {
+    name = "_amazonses.${aws_ses_domain_identity.wellcomecollection_org.id}"
+    type = "TXT"
+    records = [
```
garyttierney, line 13: Needs a `terraform fmt` I think.

garyttierney (changes requested):

infra/static/output.tf
danielgrant: `terraform fmt` is fine with it actually, or it wouldn't have passed the build! Fixed it anyway.

garyttierney (approved):

PR #24, opened 2020-12-03
Enhancing API Gateway documentation

danielgrant:
This PR includes:

- Updating the API documentation for the various components of the API.
- Exposing API Gateway's generated Swagger via an S3 / CloudFront hosted Swagger UI.


packages/shared/sierra-client/src/index.ts
```diff
@@ -1,7 +1,7 @@
-import type {AxiosInstance} from 'axios';
+import { APIResponse, errorResponse, ResponseStatus, successResponse, unhandledError } from '@weco/identity-common';
```
garyttierney, line 1: Is there an auto-formatter you're using to fix these or just IntelliJ?

danielgrant: The IntelliJ auto-formatter doesn't organise imports very well - or at all for that matter. The VSCode one is much better, so I ran that through them initially and then try and maintain that format by hand as time goes on.

packages/apps/api/src/app.ts
```diff
@@ -1,69 +1,119 @@
...
-app.post('/auth', (req, res) => validateCredentials(sierraClient, req, res));
+function createApplication(): Application {
+  const app: Application = express();
 
-app.get('/users', function (req: Request, res: Response) {
-  console.log('Processing [GET /users]...');
-  res.status(200).json([DummyUserOne, DummyUserTwo]);
-});
+  app.use(bodyParser.json());
+  app.use(bodyParser.urlencoded({ extended: true }));
```
garyttierney, line 24: You can use `app.use(...)` on your CORS calls too so you can centralize them here:

```js
const corsOptions: cors.CorsOptions = {
   allowedHeaders: ['Content-Type', 'X-API-Key'],
   methods: 'OPTIONS,POST',
   origin: process.env.API_ALLOWED_ORIGINS
};

app.use(cors(corsOptions));
```

garyttierney (changes requested):
Documentation LGTM, some comments on the API routing code.

packages/apps/api/src/app.ts
danielgrant: Yes indeed and that was how it was during the initial testing, but some of the resources accept different headers and different methods. If we used the catch-all `app.use(...)` approach, then e.g.

```
[OPTIONS] /auth
```

would return `GET, PUT, DELETE, POST` for the available methods and `X-API-Key` and `Authorization` for the acceptable methods, whereas this endpoint actually only supports `POST` and requires the `X-API-Key`. By breaking out each resource into its own CORS configuration, this allows us to specific about what methods and heads that exact resource path supports.

```
app.options('*', cors()) // include before other routes
```

This would never actually be called because the API Gateway that sits above the API Lambda doesn't have the concept of a wildcard path, so any query to any resource path that isn't explicitly created in API Gateway and subsequently reflected in this Lambda's mapping would just return a 404 by the API Gateway and wouldn't make it this far.

garyttierney (approved):

PR #26, opened 2020-12-08
Implements backend API support for user creation

danielgrant:
Implements the `[POST] /users` API operation. This involves creation of a Patron record in Sierra and a corresponding Auth0 user, the two of which are linked together via their corresponding identifiers.

### DIFF
--- a/packages/shared/sierra-client/src/patron.ts
+++ b/packages/shared/sierra-client/src/patron.ts
@@ -85,7 +85,27 @@ export function toCreatePatron(title: string, firstName: string, lastName: strin
         label: "USER CAT.",
         value: "13"
       },
-    }
+    },
+    varFields: [
+      {
+        fieldTag: "n",
+        marcTag: "100",
+        subfields: [
+          {
+            tag: "a",
+            content: lastName
+          },
+          {
+            tag: "c",
+            content: title
+          },
+          {
+            tag: "b",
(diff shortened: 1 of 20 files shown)


packages/shared/sierra-client/src/patron.ts
```diff
@@ -74,25 +76,92 @@ function getPatronNameNonMarc(content: string): PatronName {
...
+
+interface VarField {
+  fieldTag: string,
+  content?: string,
+  subfields?: SubField[]
+}
+
+interface SubField {
+  tag: string,
+  content: string
```
garyttierney, line 166: Are these types ever used? I can't find any references to them.

danielgrant: [Here](https://github.com/wellcomecollection/identity/blob/feat/account_registration_backend_support/packages/shared/sierra-client/src/patron.ts#L1), [here](https://github.com/wellcomecollection/identity/blob/feat/account_registration_backend_support/packages/shared/sierra-client/src/patron.ts#L10) and [here](https://github.com/wellcomecollection/identity/blob/feat/account_registration_backend_support/packages/shared/sierra-client/src/patron.ts#L33).

garyttierney (approved):

garyttierney (approved):

PR #28, opened 2020-12-09
Implements unit testing in Typescript modules

danielgrant:


stephenwf:
In terms of simplifying configuration it might be easier if we configure jest at the root of the mono-repo once and run all of the tests together. This will produce a single coverage file and still allow you to run isolated folders per-project if you need.

jest.config.base.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+    roots: [
+        "<rootDir>/src",
+        "<rootDir>/tests"
+    ],
+    transform: {
+        "^.+\\.ts$": "ts-jest"
```
stephenwf, line 7: The `testRegex` additionally looks for `.tsx` extension, we should add that to this regex too so that they are picked up and ran through `ts-jest`

jest.config.json
```diff
@@ -0,0 +1,24 @@
...
+    "moduleFileExtensions": [
+        "ts",
+        "tsx",
+        "js",
+        "jsx",
+        "json",
+        "node"
+    ],
+    "moduleNameMapper": {
+        "@weco/(.+)$": "<rootDir>packages/shared/$1/src"
```
stephenwf, line 22: In the `auth0-client` the rootDir is overridden to be `./packages/auth0-client` resulting in this mapping being: `./packages/auth0-client/packages/shared/$1/src` which does not exist.

garyttierney: Isn't `<rootDir>` here relating to the jest config, not TypeScript? It does seem to work. I'll check with the change.

garyttierney (dismissed):
Besides Stephen's comment which I'm testing separately, LGTM.

garyttierney (approved):

PR #30, opened 2020-12-10
Adds HTML for the log in and the verification email

ionab:
Note, the configuration for the forgot password page is encompassed in the html for the log in.


packages/apps/auth0-actions/src/verification-email.html
```diff
@@ -0,0 +1,64 @@
...
+            id="bodyCell"
+            style='-webkit-text-size-adjust: 100%;-ms-text-size-adjust: 100%;mso-table-lspace: 0pt;mso-table-rspace: 0pt;margin: 0;padding: 20px;font-family: "ProximaNova", sans-serif;height: 100% !important;'
+          >
+            <div class="main">
+              <p
+                style="text-align: center;-webkit-text-size-adjust: 100%;-ms-text-size-adjust: 100%; margin-bottom: 30px;"
+              >
+                <img
+                  src="https://identity-public-assets-stage.s3.eu-west-1.amazonaws.com/images/wellcomecollections-150x50.png"
+                  width="200"
```
danielgrant, line 32: If this image is to be 200 (pixels?) wide, should we not be using a logo that was rendered at that size? Stretching the 150x150 logo may cause it to lose quality.

packages/apps/auth0-actions/src/verification-email.html
```diff
@@ -0,0 +1,64 @@
...
+                  alt="Wellcome Collection Logo"
+                  style="-ms-interpolation-mode: bicubic;border: 0;height: auto;line-height: 100%;outline: none;text-decoration: none;"
+                />
+              </p>
+
+              <p>Thank you for signing up. Please verify your email address by clicking the following link:</p>
+
+              <p><a href="{{ url }}">Confirm my account</a></p>
```
danielgrant, line 40: This is not an account confirmation email, it's an email address verification email. The account is active and working regardless of whether the user chooses to click the URL here. Should the text not reflect that distinction?

packages/apps/auth0-actions/src/verification-email.html
```diff
@@ -0,0 +1,64 @@
...
+
+              <p>
+                If you are having any issues with your account, please don’t hesitate to contact us by replying to
```
danielgrant, line 43: There's actually support details configured on the tenant level for this purpose, available as {{ support_email }} and {{ support_url }} - I would suggest we reference that here instead as there's no guarantee that the address from which this email is sent will be able to receive replies.

packages/apps/auth0-actions/src/verification-email.html
```diff
@@ -0,0 +1,64 @@
...
+              <br />
+              Thanks!
+              <br />
+
+              <strong>Wellcome Collection</strong>
+
+              <br /><br />
+              <hr style="border: 2px solid #EAEEF3; border-bottom: 0; margin: 20px 0;" />
+              <p style="text-align: center;color: #A9B3BC;-webkit-text-size-adjust: 100%;-ms-text-size-adjust: 100%;">
+                If you did not make this request, please contact us by replying to this mail.
```
danielgrant, line 56: Same as above.

packages/apps/auth0-actions/src/login.html
```diff
@@ -0,0 +1,80 @@
...
+        __tenant: config.auth0Tenant,
+        __token_issuer: config.authorizationServer.issuer
+      },
+      assetsUrl:  config.assetsUrl,
+      allowedConnections: connection ? [connection] : null,
+      rememberLastLogin: !prompt,
+      language: language,
+      languageDictionary: languageDictionary,
+      theme: {
+        logo:            'https://identity-public-assets-stage.s3.eu-west-1.amazonaws.com/images/wellcomecollections-150x50.png',
```
danielgrant, line 58: Ideally we could insert a placeholder value here and inject the real value at build time.

packages/apps/auth0-actions/src/login.html
```diff
@@ -0,0 +1,80 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+  <title>Sign In with Auth0</title>
```
danielgrant, line 6: Should this not be something else? We shouldn't be making it obvious to the end user that this is an Auth0 platform, that sort of thing should be invisible.

danielgrant (changes requested):

danielgrant:
On another note - I don't think `packages/apps/auth0-actions` is the right place to store this. The scripts stored in there are Auth0 action scripts, i.e. JavaScript that is executed under certain conditions.

packages/apps/auth0-actions/src/login.html
ionab: Do you know of a way to do this @danielgrant? It's not not clear in the documentation: 
https://auth0.com/docs/libraries/lock/lock-ui-customization#theming-options

danielgrant: There's no built in way of doing it that I can see but we can template it out in the repo and render that as part of our own build process, since we know the URL of the image. Leave it for now and we can pick this up later.

danielgrant (approved):

PR #36, opened 2020-12-15
Allow user to register for an account, front end

ionab:
Allows the user to register for an account frontend - proxy implimentation.
Proxy -> API handshake to be confirmed. 

PR introduced routes for the /register and /validated in the user registration flow. 

PR also brings in to the code base, draft react display only code for other elements of the application.


packages/apps/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -9,14 +9,15 @@ import SpacingComponent from '@weco/common/views/components/SpacingComponent/Spa
...
-import { AccountValidated } from '../AccountManagement/AccountValidated';
+import { AccountValidated } from './AccountValidated';
 import { ErrorMessage } from '../Shared/ErrorMessage';
 
 const logo = 'https://identity-public-assets-stage.s3.eu-west-1.amazonaws.com/images/wellcomecollections-150x50.png';
 import styled from 'styled-components';
 
 // At least 8 characters, one uppercase, one lowercase and number
 const passwordPolicy = /(?=.{8,})(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9]).*/;
+const emailTest = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
stephenwf, line 20: Could we use an `<input type="email" />` and let the browser validate the email? I'm not sure how we can get the validity of the input to show a message, so perhaps it's not possible but there's a small risk that the regex may not work on all valid emails.

ionab: We're also using the browser input type here, just using the regex as a double check to display message to user if we flag it!

stephenwf (approved):

packages/account-management/src/frontend/AccountManagement/DeleteAccount.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import { SolidButton } from '@weco/common/views/components/ButtonSolid/ButtonSolid';
+import SpacingComponent from '@weco/common/views/components/SpacingComponent/SpacingComponent';
+
+const logo = 'https://identity-public-assets-stage.s3.eu-west-1.amazonaws.com/images/wellcomecollections-150x50.png';
```
robmarch2, line 6: Looks like this URL is just to the staging environment, which isn't a huge deal right now but we'll need to change before we deploy to prod. It's in a few other places too. Can we either stick a `TODO` before each usage, or get them all from a common place (e.g. in `config.ts`) with a `TODO` on that?

robmarch2 (approved):

PR #46, opened 2021-01-13
Integrate account mangement registration page with backend API

robmarch2:


stephenwf (approved):

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -59,8 +59,14 @@ export const Registration: React.FC = () => {
   }, []);
 
   useEffect(() => {
-    if (email !== '') setEmailValid(emailTest.test(email || ''));
-    if (email === '') setEmailValid(true);
+    if (email !== '') {
+      setEmailValid(emailTest.test(email || ''));
+      setAlreadyExists(false);
+    }
+    if (!email || email === '') {
```
ionab, line 66: Will this result in the user being trying to create an account with a blank password field?

ionab (changes requested):

packages/account-management/src/frontend/Registration/Registration.tsx
robmarch2: No, we don't actually make a post to the API unless all the fields have been entered (see line 83). This is just about making sure that the error on the email field isn't shown on initially loading the page (the initial condition on this check was `email === ''` which was evaluating false on initial page load), and getting rid of the duplicate email address error when changing the email field. Without this, if you were to use a duplicate email and common password, submit the form, then change the email address and resubmit, the initial error message about the email being duplicated would still be hanging around despite this not being the case any more.

This does remind me that I thought I'd made a similar change on the password field which no longer seems to be there after cherry picking my changes over to the latest master so I'm just going to double check that that

ionab (approved):

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -71,7 +71,13 @@ export const Registration: React.FC = () => {
...
   useEffect(() => {
     // check if password passes password policy
-    pass && !passwordPolicy.test(pass || '') ? setPassQualifies(false) : setPassQualifies(true);
+    if (pass && !passwordPolicy.test(pass || '')) {
+      setPassQualifies(false);
+      setCommonPassword(false);
+    } else {
+      setPassQualifies(true);
+      setCommonPassword(false);
+    }
```
ajrussellaudio, line 80: Can we simplify this?

```suggestion
    setPassQualifies(pass && passwordPolicy.test(pass))
    setCommonPassword(false)
```

ajrussellaudio (changes requested):

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -70,14 +70,8 @@ export const Registration: React.FC = () => {
...
   useEffect(() => {
-    // check if password passes password policy
-    if (pass && !passwordPolicy.test(pass || '')) {
-      setPassQualifies(false);
-      setCommonPassword(false);
-    } else {
-      setPassQualifies(true);
-      setCommonPassword(false);
-    }
+    setPassQualifies(!pass || passwordPolicy.test(pass || ''));
```
ajrussellaudio, line 73: On second thought, we can just do:

```suggestion
    setPassQualifies(passwordPolicy.test(pass));
```

`.test` can handle `undefined` and empty strings nae bother:

```
> const regex = /[A-Z]+/
undefined
> regex.test()
false
> regex.test('')
false
> regex.test('abc')
false
> regex.test('ABC')
true
```

ajrussellaudio (changes requested):

packages/account-management/src/frontend/Registration/Registration.tsx
robmarch2: The issue here is about making sure we don't flag the password as invalid before the user has entered anything. If we make that change, the invalid password error message will be displayed from the moment the page loads

ajrussellaudio: That makes sense. In that case, this solution is a good one :+1:

ajrussellaudio (dismissed):

ionab (approved):

ionab (approved):

ajrussellaudio (approved):

PR #49, opened 2021-01-14
Account Management page

ajrussellaudio:
This PR:

* Reworks the Wellcome `TabNav` component to work with React Router (previously depended on Next.js' router)
* Makes adjustments to the `AccountManagement` page in line with latest UX designs
* Adds a `/account` route to the above page


packages/account-management/src/frontend/index.tsx
```diff
@@ -13,12 +14,21 @@ import '@weco/common/styles/styleguide.scss';
...
     padding: 42px;
   }
+=======
+  width: 70%;
+  margin: auto;
+  background-color: #f0ede3;
+  padding: 0 42px;
+  max-height: 100%;
+  overflow: auto;
+>>>>>>> Adjust ProfileForm in line with UX design
```
stephenwf, line 31: Possibly a merge conflict in the CSS?

packages/account-management/src/frontend/AccountManagement/TabNav.tsx
```diff
@@ -0,0 +1,108 @@
+// Converted from flow to TSX element
```
ionab, line 1: My suggestion as we convert weco components and introduce them to the codebase would be to keep them in a seperate folder for maintainability for tracking which components we've had to do this with?

ionab (changes requested):

packages/account-management/src/frontend/index.tsx
ajrussellaudio: Yikes. Forgot to hit Save 🤦

packages/account-management/src/frontend/AccountManagement/TabNav.tsx
ajrussellaudio: Good idea. Is the `frontend/Shared` folder a good place for it to live? It looks like `frontend/Shared/CheckBoxLabel.tsx` was also adapted from a WeCo component, based on the comments there.

ionab: Yeah I'd suggest maybe both of them go into a subfolder of shared.... or both out to their own folder not nested in shared. I don't think it really matters as long as we can keep good track of them and it's clear to weco what we've done.

ajrussellaudio:
`tsc` was giving an error from one of the Storybook files:

```
src/stories/accountAdmin.stories.tsx:14:7 - error TS2322: Type '{ existingTitle: string; existingFirstName: string; existingLastName: string; existingEmail: string; }' is not assignable to type 'IntrinsicAttributes & UserInfo & { children?: ReactNode; }'.
  Property 'existingTitle' does not exist on type 'IntrinsicAttributes & UserInfo & { children?: ReactNode; }'.

14       existingTitle={text('Title', 'Lord')}
         ~~~~~~~~~~~~~
```

Storybook isn't working in this branch, but the last commit fixes the `tsc` warning.

ionab (approved):

PR #51, opened 2021-01-14
Redirect to login

ionab:
Adds redirect to the `/login` page from the `/validated `route.

### DIFF
--- a/packages/account-management/src/frontend/Registration/AccountValidated.tsx
+++ b/packages/account-management/src/frontend/Registration/AccountValidated.tsx
@@ -29,7 +29,6 @@ export const AccountValidated = () => {
   const history = useHistory();
 
   const redirectToLogin = () => {
-    // Redirect to login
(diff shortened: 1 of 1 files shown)


packages/account-management/src/frontend/Registration/AccountValidated.tsx
```diff
@@ -18,25 +18,32 @@ const LogoContainer = styled.div`
...
+    flex-direction: column;
+    align-items: center;
+  }
+`;
 
 export const AccountValidated = () => {
+  const history = useHistory();
+
+  const redirectToLogin = () => {
+    // Redirect to login
```
ajrussellaudio, line 32: Can we live without this comment?

ajrussellaudio (changes requested):

ajrussellaudio (approved):

ajrussellaudio (approved):

PR #52, opened 2021-01-14
[ADR-0004] Auth Strategy: Passport vs OpenID Connect

ajrussellaudio:


danielgrant (changes requested):
I'll let @stephenwf @ajrussellaudio @ionab make the decision on this one, but it sounds like sticking with passport-auth0 is the path of least resistance here. We already have it integrated and Auth0 will continue to support the library for the time being, so I don't see any pressing need to switch to express-openid-connect unless someone else can provide one.

ionab (dismissed):
Looks reasonable to me, I think the decision should be to stick with Passport-Auth0, as we are using KOA.

docs/adr/0004-auth-strategy--passport-vs-openid.md
```diff
@@ -0,0 +1,40 @@
...
+### In favour of Passport
+
+- The work to implement user authentication in this project is more or less done already, and meets the customer's requirements.
+- The app has no dependency on Auth0's Passport strategy for which development is stopping. Passport integration is achieved with [koa-passport](https://github.com/rkusa/koa-passport).
+- The Node.js server uses the Koa framework, but the OpenID Connect SDK is written for Express. Express middleware is not directly compatible with Koa, although solutions to achieve compatibility do exist.
+
+### In favour of OpenID Connect SDK
+
+- Based on their [examples](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md), Auth0's OpenID Connect SDK seems to _vastly_ reduce the amount of code necessary to achieve the standard aims of user authentication.
+- The koa-passport repo has not been updated since July 2019, and the package not published since April 2019. As it is a wrapper around the main Passport library, it is not clear if there are security implications of such a long time without updates.
```
stephenwf, line 36: We can, if required, update the passport dependency independently.

docs/adr/0004-auth-strategy--passport-vs-openid.md
```diff
@@ -0,0 +1,40 @@
...
+## Considered Options
+
+- Continue using Passport
+- Drop Passport in favour of OpenID Connect with the Auth0 SDK
+
+### In favour of Passport
+
+- The work to implement user authentication in this project is more or less done already, and meets the customer's requirements.
+- The app has no dependency on Auth0's Passport strategy for which development is stopping. Passport integration is achieved with [koa-passport](https://github.com/rkusa/koa-passport).
+- The Node.js server uses the Koa framework, but the OpenID Connect SDK is written for Express. Express middleware is not directly compatible with Koa, although solutions to achieve compatibility do exist.
```
stephenwf, line 31: This is a small one, but we do use the passport to enable local development, which would be lost or require re-work.

stephenwf (approved):
I agree, there is not a considerable downside to sticking with koa-passport 👍

stephenwf (approved):

danielgrant (dismissed):

danielgrant (approved):

PR #53, opened 2021-01-14
Registration feedback

ionab:
Impliments the feedback from Gareth: 

> Form section labels, lowercase d for details: ‘Personal details’ and ‘Login details’
> We have a different treatment for error messaging (see attached file in case you aren’t aware)
> The bullet list with password hints could be moved closer to the field to give it a stronger association
> Check box would need a white background for AA compliancy


packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -104,18 +106,34 @@ export const Registration: React.FC = () => {
...
         <img src={logo} alt="Wellcome Collection Logo" />
       </LogoContainer>
-      {errorOccured ? (
-        <ErrorMessage>
-          Something went wrong.
-        </ErrorMessage>
-      ) : (
-        <></>
-      )}
+      {errorOccured ? <ErrorMessage>Something went wrong.</ErrorMessage> : <></>}
```
ajrussellaudio, line 136:
```suggestion
      {errorOccured && <ErrorMessage>Something went wrong.</ErrorMessage>}
```

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -144,43 +165,34 @@ export const Registration: React.FC = () => {
...
-            ) : (
+            <PasswordInput
+              value={pass}
+              setValue={(value: string) => setPass(value)}
+              pattern={passwordPolicy}
+              isValid={!commonPassword && passQualifies}
+              showValidity={!!valid}
+              errorMessage={passwordErrorMessage()}
+            />
+            {!passQualifies ? null : (
```
ajrussellaudio, line 195: We could make this simpler:

```ts
{passQualifies && (
  <ul>
    // ...
  </ul>
)}
```

...but do we really only want to show the list of rules when the password qualifies?

packages/account-management/src/frontend/Shared/PasswordInput.tsx
```diff
@@ -12,13 +12,18 @@ const IconWrapper = styled.div`
...
+  margin-top: -35px;
 `;
 
 type PasswordInputProps = {
   value: string | undefined;
   setValue: (value: string) => void;
   pattern?: RegExp;
+  showValidity?: boolean;
+  isValid?: boolean;
+  errorMessage?: any;
```
ajrussellaudio, line 26: Can this be a `string` or `ReactNode`?

packages/account-management/src/frontend/Shared/PasswordInput.tsx
```diff
@@ -28,6 +33,9 @@ export const PasswordInput: React.FC<PasswordInputProps> = (props) => {
 
   return (
     <PasswordFieldWrapper>
+      <IconWrapper>
+        {!showPassword ? <EyeIcon onClick={toggleShowPassword} /> : <AllyVisual onClick={toggleShowPassword} />}
```
ajrussellaudio, line 37: I think I committed this line earlier today, but now I see it again, we might as well flip the logic for readability:

```suggestion
        {showPassword ? <AllyVisual onClick={toggleShowPassword} /> : <EyeIcon onClick={toggleShowPassword} />}
```

My bad 😬

ajrussellaudio (changes requested):

robmarch2 (changes requested):
Some things aren't quite right with the error handling on this. I'm not seeing any error messages on the email field prior to the password field passing the rules, and I'm also not seeing any errors on the password field indicating that it's not passing the rules. The list of rules are also hidden unless the password field is empty, or matches the rules. I've also noticed an issue with duplicate account error message:

image

packages/account-management/src/frontend/Registration/Registration.tsx
ionab: The password error message spells it out which is why I've hidden it?

packages/account-management/src/frontend/WellcomeComponents/TextInput.tsx
```diff
@@ -0,0 +1,244 @@
...
+  isValid?: boolean,
+  setIsValid?: (value: boolean) => void,
+  showValidity?: boolean,
+  setShowValidity?: (value: boolean) => void,
+  autoFocus?: boolean,
+  big?: boolean,
+  ariaLabel?: string,
+};
+
+// $FlowFixMe (forwardRef)
```
ajrussellaudio, line 163: Should this be removed?

packages/account-management/src/frontend/WellcomeComponents/TextInput.tsx
```diff
@@ -0,0 +1,244 @@
...
+            aria-label={ariaLabel}
+          />
+          {isValid && showValidity && (
+            <TextInputCheckmark>
+              <Icon name={`check`} extraClasses={`icon--green`} />
+            </TextInputCheckmark>
+          )}
+        </TextInputWrap>
+        {errorMessage && !isValid && showValidity && (
+          <TextInputErrorMessage dangerouslySetInnerHTML={{ __html: errorMessage }}/>
```
ajrussellaudio, line 235: Could we adapt this component to accept some JSX as an `errorMessage` instead of an HTML string? Something like:

```tsx
{errorMessage && !isValid && showValidity && (
  <TextInputErrorMessage>
    {errorMessage}
  </TextInputErrorMessage>
)}
```

```ts
// /src/frontend/Registration/Registration.tsx

const emailErrorMessage = () => {
  if (alreadyExists) {
    return (
      <>
        This account already exists. You can try to
        <a href="/">login</a>
      </>
  } else // ...

// ...

<TextInput
  errorMessage={emailErrorMessage()}
/>
```

ajrussellaudio (changes requested):

packages/account-management/src/frontend/Registration/Registration.tsx
ajrussellaudio: It's a UX call, but I think it's ok to do both. The user should see the rules before they start typing, and if there's an error it should explain why the password failed.

ionab:
> Some things aren't quite right with the error handling on this. I'm not seeing any error messages on the email field prior to the password field passing the rules, and I'm also not seeing any errors on the password field indicating that it's not passing the rules. The list of rules are also hidden unless the password field is empty, or matches the rules. I've also noticed an issue with duplicate account error message:
> 
> image

Good spot Rob, I've brought the weco TextInput across and converted in into a typescipt one for use here so we can nest react nodes in the messaging.

packages/account-management/src/frontend/Registration/Registration.tsx
ionab: I've taken out the condition here for now, as wireframes don't show either way. If there is a requirement to hide it when there is an error message shown, it's straightforward enough to add it in!

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -28,7 +28,7 @@ export const Registration: React.FC = () => {
   const [consent, setConsent] = useState(false);
   const [alreadyExists, setAlreadyExists] = useState(true);
   const [passQualifies, setPassQualifies] = useState(true);
```
ajrussellaudio, line 30: While we're here, should this also initialise as `false`?

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -147,8 +152,8 @@ export const Registration: React.FC = () => {
               label="First name"
               value={firstName}
               setValue={(value: string) => setFirstName(value)}
-              isValid={!!(firstName && firstName !== '')}
-              showValidity={!!valid}
+              isValid={firstName !== ''}
+              showValidity={!!firstName && firstName !== ''}
```
ajrussellaudio, line 156: Could we use a variable to avoid repetition here?

```ts
const isFirstNameValid = Boolean(firstName && firstName !== '');

// ...

  isValid={isFirstNameValid}
  showValidity={isFirstNameValid}
```

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -160,8 +165,8 @@ export const Registration: React.FC = () => {
               label="Last name"
               value={lastName}
               setValue={(value: string) => setLastName(value)}
-              isValid={!!(lastName || lastName === '')}
-              showValidity={!!valid}
+              isValid={lastName !== ''}
+              showValidity={!!lastName && lastName !== ''}
```
ajrussellaudio, line 169: Similar to `firstName` above, could we use a variable to avoid repetition here?

ajrussellaudio (changes requested):
Nothing major here, just a few wee things

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -64,7 +53,7 @@ export const Registration: React.FC = () => {
   }, [pass]);
 
   useEffect(() => {
-    setValid(firstName && lastName && email && passwordPolicy.test(pass || ''));
+    setValid(!!(firstName && lastName && email && passwordPolicy.test(pass || '')));
```
robmarch2, line 56: What's this `!!` operator? It's in a few places in here but I've had a google but the only thing I can understand it as being is `not not` which would be redundant...

ajrussellaudio: `not not` is right, but it's a quick and dirty way to convert something to a boolean, equivalent to `Boolean(firstName && lastName && ...`

packages/account-management/src/frontend/Registration/Registration.tsx
```diff
@@ -56,6 +56,9 @@ export const Registration: React.FC = () => {
     setValid(!!(firstName && lastName && email && passwordPolicy.test(pass || '')));
   }, [firstName, lastName, email, pass, consent]);
 
+  const isFirstNameValid = Boolean(firstName && firstName !== '');
+  const isLastNameValid = Boolean(lastName && lastName !== '');
```
ajrussellaudio, line 60: Following the conversation on Friday, we don't actually need `Boolean` or `!!` here!

The right side of the `&&` evaluates to a boolean already. So we either get `false` because the name isn't set, or `false` because it's an empty string, or `true` when it's not empty.

ajrussellaudio (changes requested):

packages/account-management/src/frontend/Registration/Registration.tsx
ionab: @ajrussellaudio taking out `Boolean()` or `!!` here results in the following error on the type check: 

```
Type 'boolean | "" | undefined' is not assignable to type 'boolean | undefined'.
  Type '""' is not assignable to type 'boolean | undefined'.
```

ajrussellaudio: Yikes. How about:

```ts
const isFirstNameValid = firstName && (firstName !== '');
```

ionab: Yup, this results in the same type error!

ajrussellaudio: Dammit, TypeScript! 😅 

If that's the case, your solution is the right one 👍

ajrussellaudio (approved):

robmarch2 (approved):

PR #54, opened 2021-01-15
Add endpoint to verify user credentials

garyttierney:


packages/apps/api/src/app.ts
```diff
@@ -65,6 +65,7 @@ function registerUsersUserIdResource(app: Application): void {
   app.options('/users/:user_id', cors(corsOptions));
   app.get('/users/:user_id', cors(corsOptions), (request: Request, response: Response) => getUser(sierraClient, auth0Client, request, response));
   app.put('/users/:user_id', cors(corsOptions), (request: Request, response: Response) => updateUser(sierraClient, auth0Client, request, response));
+  app.post('/users/:user_id/auth', cors(corsOptions), (request: Request, response: Response) => authenticateUser(auth0Client, request, response));
```
danielgrant, line 68: I wonder if this is the correct path for this?

Looking at `authenticateUser` in `user.ts` below, it looks like we expect both the user ID to be provided as a path parameter, and the email address and password as request parameters. The issue I think we _may_ run into in the future is if ourselves or someone else wants to be able to test a users credentials (email address and password) before they know what the users ID is. In this case, they won't be able to populate the `:user_id` path parameter. In general I don't think it makes sense to have to provide both user ID and email address anyway.

We have an existing endpoint `[POST] /auth` that currently tests Sierra credentials. This was originally going to be used by the Auth0 action scripts, but we've changed that so Auth0 goes directly to Sierra instead and as a result that endpoint is now unused.

How would you feel about switching the path here so that the Auth0 credential testing makes use of the existing [`POST] /auth` endpoint instead?

packages/apps/api/src/handlers/user.ts
```diff
@@ -31,6 +31,59 @@ export async function getUser(sierraClient: SierraClient, auth0Client: Auth0Clie
...
 
+export async function authenticateUser(auth0Client: Auth0Client, request: Request, response: Response) {
+  const email: string = request.body.email;
+  const password: string = request.body.password;
+
+  if (!isNonBlank(email) || !isNonBlank(password)) {
+    response.status(400).json(toMessage("All fields must be provided and non-blank"));
+  }
+
+  const validationResult = await auth0Client.validateUserCredentials(email, password);
```
danielgrant, line 42: Should we not be checking for 404 here too, i.e. given email address not found? Does the Auth0 API return a specific error code that we can catch and communicate back to the caller if the account is blocked?

danielgrant (changes requested):

packages/apps/api/src/app.ts
danielgrant: As a note here too - the handler here would be registered in its own function since it's not a method operating on the `/user/:user_id` resource - see below for examples.

garyttierney: Ah, I didn't realize there were other user endpoints already that weren't under here. I'll move that out.

+1 on updating to use /auth, doing it by userId doesn't really make sense anyway since the parameter is never actually used.

packages/apps/api/src/handlers/user.ts
garyttierney: Not in the status code, no. The OAuth2 specification does say that a human-readable error should be present in `error_description` on the response, but the status code will only ever be 200, 400, or 403.

garyttierney: The docs do say that the message can be something like "MFA Auth required", but it can also show things like:

```json
{  "error_description":"Authorization pending: please repeat the request in a few seconds."  }
```

danielgrant: Could we look the given email address up first, and if it isn't found, return a 404? Then if it does exist, proceed with the authentication request?

garyttierney: Yeah, that should be fine to do in a private API.

garyttierney: It does blow up this request by an extra 3 HTTP calls, which I think we should be conscious of. It might be better to change it later to get the error from whatever the token endpoint returns.

danielgrant: What are the 3 calls? In my head it goes something like this:

1. Client calls API with email address and password
2. API calls Auth0 Managament API to validate email address exists
3. API calls Auth0 Authentication API to validate email address and password

If (2) or (3) fail, we return the appropriate status code back to the client immediately. To me checking that the email address exists and returning a `404` back to the caller if it doesn't adds 1 extra call.

Am I missing something?

garyttierney: Ah it's 2 extra calls, not 3. I misread this bit:

```ts
    return axios.post(this.apiRoot + '/oauth/token', { // (1)
      client_id: this.clientId,
      client_secret: this.clientSecret,
      audience: this.apiAudience,
      grant_type: 'client_credentials'
    }, {
      validateStatus: status => status === 200
    }).then(response => {
      return axios.create({                           // misread as (2)
        baseURL: this.apiRoot + '/api/v2',
        headers: {
          Authorization: 'Bearer ' + response.data.access_token
        }
      });
    });
```

We weren't using a machine to machine token before, so we need to get that before we can validate the user exists.

packages/shared/auth0-client/src/index.ts
```diff
@@ -103,6 +126,24 @@ export default class Auth0Client {
     });
   }
 
+  async validateUserCredentials(username: string, password: string) : Promise<APIResponse<boolean>> {
+    const userFetchResult = await this.getUserByEmail(username);
+
+    if (userFetchResult.status != ResponseStatus.Success) {
```
danielgrant, line 132: How do you feel about doing this check as part of `packages/apps/api/src/handlers/auth.ts#validateCredentials`? That way we keep this particular call specific to its purpose, and the Lambda Function can concern itself with checking if the user exists?

packages/shared/auth0-client/src/index.ts
```diff
@@ -16,6 +20,25 @@ export default class Auth0Client {
...
+
+    return httpClient.delete(`/users/auth0|p${userId}`, { validateStatus: responseCodeIs(204), })
+      .then(_ => {
+        return successResponse(true);
+      })
+      .catch((error: AxiosError) => {
+        if (error.response) {
+          switch (error.response.status) {
+            case 404:
+              return errorResponse(`User with id "auth0|p${userId}" not found`, ResponseStatus.NotFound, error);
```
danielgrant, line 34: This error message exposes the inner working of Auth0 to the client, i.e. the `auth0|p` ID prefix. I've made a concious effort to avoid leaking this stuff outside of this class, so everywhere above here should just deal with the numerical user ID.

Where we report the similar error message elsewhere, we use:

```
'Auth0 user with ID [' + userId + '] not found'
```

where `userId` is the numerical ID of the user that aligns with Sierra.

packages/shared/auth0-client/src/index.ts
```diff
@@ -1,7 +1,11 @@
 import { APIResponse, errorResponse, ResponseStatus, successResponse, unhandledError } from '@weco/identity-common';
-import axios, { AxiosInstance } from 'axios';
+import axios, { AxiosError, AxiosInstance } from 'axios';
 import { Auth0Profile, Auth0UserInfo, toAuth0Profile, toAuth0UserInfo } from './auth0';
 
+function responseCodeIs(responseCode: number) {
```
danielgrant, line 5: Something for `shared/identity-common` perhaps, so we can reuse it both here and in `sierra-client`?

packages/shared/auth0-client/src/index.ts
```diff
@@ -181,6 +222,26 @@ export default class Auth0Client {
     });
   }
 
+  private getInstanceWithCredentials(username: string, password: string): Promise<AxiosInstance> {
+    return axios.post(this.apiRoot + '/oauth/token', {
+      client_id: this.clientId,
+      client_secret: this.clientSecret,
+      audience: this.apiAudience,
+      grant_type: 'password',
+      username,
```
danielgrant, line 231: I assume that if we don't provide a key here, it'll just use the name of the variable...?

danielgrant (changes requested):

packages/shared/auth0-client/src/index.ts
garyttierney: Yeah, `{ username }` is syntax sugar for `{ username: username }`

packages/shared/auth0-client/src/index.ts
garyttierney: Yep, actually forgot to use it in my second commit. Going to update that and will move it.

packages/shared/auth0-client/src/index.ts
garyttierney: yeah, I wasn't really sure about this initially, but keeping /auth strictly to auth seems nicer. It makes me less nervous about user enumeration even though it's a private API endpoint.

packages/apps/api/src/handlers/auth.ts
```diff
@@ -11,6 +11,11 @@ export async function validateCredentials(auth0Client: Auth0Client, request: Req
     response.status(400).json(toMessage("All fields must be provided and non-blank"));
   }
 
+  const userFetchResult = await auth0Client.getUserByEmail(email);
+  if (userFetchResult.status != ResponseStatus.Success) {
+    return errorResponse(userFetchResult.message, userFetchResult.status);
```
danielgrant, line 16: I think this is a copy-paste error? `return errorResponse` returns an error object. We want to do something like `response.status(404).json(userFetchResult.message)` here I think?

danielgrant (changes requested):

danielgrant (approved):

packages/apps/api/src/handlers/auth.ts
```diff
@@ -1,31 +1,26 @@
...
-export async function validateCredentials(sierraClient: SierraClient, request: Request, response: Response): Promise<void> {
-
-  const email = request.body.email;
-  const password = request.body.password;
+export async function validateCredentials(auth0Client: Auth0Client, request: Request, response: Response) {
+  const email: string = request.body.email;
+  const password: string = request.body.password;
 
   if (!isNonBlank(email) || !isNonBlank(password)) {
     response.status(400).json(toMessage("All fields must be provided and non-blank"));
```
danielgrant, line 11: Just spotted this - I think we need to `return;` out of here?

danielgrant (changes requested):

garyttierney:
Closing, but keeping this branch open for the deletion code.

PR #59, opened 2021-01-18
Add Current Password verification field to Change Password form

ajrussellaudio:
The validation logic here seems ok to me, but the UX isn't great as the new password validation happens before the user has started typing.

The logic behind showing that "not valid" message is quite fiddly, and I'd rather write some unit tests than fiddle with it any further. Installing and configuring a UI testing framework isn't a trivial piece of work, and it shouldn't hold up this ticket. This PR meets the requirements as defined.

### DIFF
--- a/packages/account-management/src/frontend/AccountManagement/PasswordForm.tsx
+++ b/packages/account-management/src/frontend/AccountManagement/PasswordForm.tsx
@@ -9,6 +9,7 @@ import { PasswordInput } from '../Shared/PasswordInput';
 const passwordPolicy = /(?=.{8,})(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9]).*/;
 
 export const PasswordForm: React.FC = () => {
+  const [currentPassword, setCurrentPassword] = useState<string>('');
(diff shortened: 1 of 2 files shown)


ionab (changes requested):
Can we show the password policy on the form below the type new password? (See bullet points on registration page). Or double check with Ville that this isn't required?
I also understood the requirement to be that a user must confirm their current password to be able to update their password... See JIRA ticket WS-31.

ajrussellaudio:
Thanks @ionab. The bullet points were an oversight, now fixed.

I'll need to check with UX on those requirements. I'd been working on the assumption that we'd send all three fields to the API at the same time, but you're absolutely right, that's not what the requirements state. Checking a password as the user types doesn't seem great for performance or security.

ajrussellaudio:
@ionab During discussion of the above issue, it was pointed out that there is the opportunity to verify the password server-side before making the call to the 'update password' endpoint. That would probably take place in `src/routes/api/update-password.ts`, but should be done as part of WS-101 and not this ticket. As per the wireframes, all three fields in one screen is what we're looking for here.

ionab (approved):

PR #61, opened 2021-01-19
Add an endpoint to test user credentials

garyttierney:


packages/apps/api/src/handlers/auth.ts
```diff
@@ -1,31 +1,27 @@
...
-      response.status(200).end();
-    } else if (sierraValidate.status === ResponseStatus.InvalidCredentials) {
-      response.status(401).json(toMessage(sierraValidate.message));
-    } else {
-      response.status(500).json(toMessage(sierraValidate.message));
-    }
-  } else if (sierraGet.status === ResponseStatus.NotFound) {
-    response.status(404).json(toMessage(sierraGet.message))
+  const validationResult = await auth0Client.validateUserCredentials(email, password);
+  if (validationResult.status == ResponseStatus.Success) {
```
danielgrant, line 22: I think we should be a bit more specific here with the error handling.

The call to `validateUserCredentials(email, password)` should return either `200`, `401` or `500` depending on whether it was successful, the authentication attempt was explicitly denied, or some other unhandled error occurred (could be errors with the Auth0 API). These should be checked for and the respective error return.

garyttierney:
Ah, made a mistake here. Removed the Auth0Client function instead of the handler.

packages/shared/auth0-client/src/index.ts
```diff
@@ -103,6 +103,18 @@ export default class Auth0Client {
     });
   }
 
+  async validateUserCredentials(username: string, password: string) : Promise<APIResponse<boolean>> {
+    return this.getInstanceWithCredentials(username, password)
+      .then(_ => successResponse(true))
+      .catch(error => {
```
danielgrant, line 109: Similar to the other functions in here, I think we should be checking for specific error codes rather than just catching any error with has a response has an invalid credentials.

For example, if invalid credentials are provided then we'll get a 401 back - I think we should specifically check for this and return `InvalidCredentials`, otherwise return a generic error.

danielgrant (changes requested):
It looks like the delete endpoint tests are still hanging around too, causing the build to fail.

packages/shared/auth0-client/src/index.ts
```diff
@@ -181,6 +193,26 @@ export default class Auth0Client {
     });
   }
 
+  private getInstanceWithCredentials(username: string, password: string): Promise<AxiosInstance> {
```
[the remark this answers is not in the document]
garyttierney: I've called it `email` elsewhere (i.e. the API client) and `username` here because we're in generic auth0-client land at this point, which doesn't/shouldn't know about business decisions that sit outside of it. 

The field you send to auth0 is `username` even if only emails are enabled.

packages/shared/auth0-client/src/index.ts
garyttierney: Huh, not sure what happened here. I thought I did check for the 401. IIRC 401 is the only error that can be returned here though, so it should only be a 401 or generic unhandled error I think.

packages/shared/auth0-client/src/index.ts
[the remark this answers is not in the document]
danielgrant: Fair enough.

packages/shared/auth0-client/src/index.ts
danielgrant: Yeah AFAICT `401` is the only non-`200` response code that can be returned, but if we catch any error and return it as an `InvalidCredentials` back to the caller then _if_ other response codes are available / introduced, then that will start returning as an `InvalidCredentials` too.

danielgrant (approved):

PR #64, opened 2021-01-20
Custom error page

ionab:
This PR adds a custom error page that will display the error message as is passed to it with the query param `error_description` as set out in https://auth0.com/docs/universal-login/custom-error-pages. 

This page is intentionally styled in a similar manner to the auth0 pages as per Wellcome requirements, explaining the introduction of the PageWrapper introduced to the other routes.


packages/account-management/src/frontend/AccountManagement/AccountManagement.tsx
```diff
@@ -6,6 +6,7 @@ import SpacingComponent from '@weco/common/views/components/SpacingComponent/Spa
 import { ProfileForm } from './ProfileForm';
 import { PasswordForm } from './PasswordForm';
 import { LogoContainer } from '../Shared/LogoContainer';
+import { PageWrapper } from '..//Shared/PageWrapper';
```
ajrussellaudio, line 9: Is there a typo in the path?

```suggestion
import { PageWrapper } from '../Shared/PageWrapper';
```

packages/account-management/src/hooks/use-location-query.tsx
```diff
@@ -0,0 +1,10 @@
+import { useLocation } from 'react-router-dom';
+import { useMemo } from 'react';
+import { parse } from 'query-string';
+
+export function useLocationQuery<T = any>() {
```
ajrussellaudio, line 5: The return type is missing here. Since it's returning a function `as T` the return type should probably be `T`:

```suggestion
export function useLocationQuery<T = any>(): T {
```

packages/account-management/src/hooks/use-location-query.tsx
```diff
@@ -0,0 +1,10 @@
+import { useLocation } from 'react-router-dom';
```
ajrussellaudio, line 1: This file (and other hooks) should probably live in the `frontend` folder. The Node server has no use for React hooks.

packages/account-management/src/frontend/Shared/ErrorPage.tsx
```diff
@@ -3,49 +3,66 @@ import SpacingComponent from '@weco/common/views/components/SpacingComponent/Spa
...
-  const redirectToWellcomeHelpPage = () => {
-    // Is it possible to configure the support url through auth0.
-    history.push('/support');
-  };
+const ErrorMessage = styled.p`
+  text-align: center;
+`;
 
+export const ErrorPage = () => {
+  const { error_description } = useLocationQuery();
```
ajrussellaudio, line 46: `useLocationQuery` can take a generic type parameter. Should we provide one here?

```ts
type ErrorQueryParams = {
  error_description: string | undefined;
}

// ...

const { error_description } = useLocationQuery<ErrorQueryParams>();
```

I haven't tested this, and there's probably a better name than `ErrorQueryParams`

ajrussellaudio (changes requested):

ajrussellaudio (approved):
Nicely done :+1:

PR #66, opened 2021-01-21
Implement the search API endpoint to facilitate searching of Auth0 users

danielgrant:
This PR additionally fixes an issue with Auth0 Terraform which causes the Auth0 action scripts to reset when the connection is modified.

### DIFF
--- a/packages/apps/api/src/handlers/user.ts
+++ b/packages/apps/api/src/handlers/user.ts
@@ -221,9 +221,9 @@ export async function searchUsers(auth0Client: Auth0Client, request: Request, re
 
   const page: number = Number(request.params.page);
   const pageSize: number = Number(request.params.pageSize);
-  const sort: string = request.body.sort;
+  const sort: string = request.params.sort;
   const sortDir: number = Number(request.params.sortDir);
-  const query: string = request.body.query;
+  const query: string = request.params.query;
   if (isNaN(page) || isNaN(pageSize) || !isNonBlank(query) || !isNonBlank(sort) || isNaN(sortDir)) {
     response.status(400).json(toMessage("All fields must be provided and non-blank"));
     return;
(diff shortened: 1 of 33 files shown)


packages/apps/api/src/handlers/user.ts
```diff
@@ -216,3 +216,40 @@ export async function changePassword(sierraClient: SierraClient, auth0Client: Au
...
+    return;
+  }
+
+  if (!Auth0SearchSortFields.get(sort)) {
+    response.status(400).json(toMessage("Invalid sort field, one of [" + Array.from(Auth0SearchSortFields.keys()) + "] must be provided"));
+    return;
+  }
+
+  if (sortDir !== 1 && sortDir !== -1) {
+    response.status(400).json(toMessage('\'sortDir\' must be \'1\' or \'-1\' to sort ascending and descending respectively'))
```
garyttierney, line 238: You switched from double quotes to single quotes here, which is odd since it means you now have to escape the single quotes that are part of the string.

packages/shared/auth0-client/src/index.ts
```diff
@@ -177,6 +192,37 @@ export default class Auth0Client {
...
+  async searchUsers(page: number, pageSize: number, sort: string, sortDir: number, query: string): Promise<APIResponse<Auth0SearchResults>> {
+    return this.getMachineToMachineInstance().then(instance => {
+      return instance.get('/users', {
+        params: {
+          page: page,
+          per_page: pageSize,
+          include_totals: true,
+          sort: Auth0SearchSortFields.get(sort) + ':' + sortDir,
+          connection: 'Sierra-Connection',
+          q: query.split(' ').map(token => 'name:*' + token + '* OR email:*' + token + '*').join(' '),
```
garyttierney, line 204: Can we break out this query generation to a separate function so we can unit test that the correct Lucene query is generated for a given set of search params?

garyttierney (approved):
2 nits, otherwise LGTM

packages/shared/auth0-client/src/index.ts
danielgrant: Good shout, done.

packages/apps/api/src/handlers/user.ts
danielgrant: I've been trying to consistently to use single quotes in the TypeScript, even if that means escaping embedded quotes in some places. I've gone through and switched everything to single quotes where I overlooked this before.

garyttierney (approved):
LGTM

packages/apps/api/src/handlers/user.ts
garyttierney: Cool, don't mind the quote style particularly, just noticed that it had switched midway.

PR #69, opened 2021-01-21
Include access token in API calls

robmarch2:


ionab (approved):

packages/account-management/src/utility/api-caller.ts
```diff
@@ -8,20 +9,26 @@ const instance: AxiosInstance = axios.create({
   },
 });
 
-export async function callRemoteApi(method: Method, url: string, body?: any): Promise<AxiosResponse> {
+export async function callRemoteApi(method: Method, url: string, contextState: ApplicationState, body?: any, authenticate: boolean = true): Promise<AxiosResponse> {
+  let headers = instance.defaults.headers;
+  if (authenticate) {
+    headers = { ... headers, 'Authorization': 'Bearer ' + contextState.user.accessToken};
+  }
```
ajrussellaudio, line 16: Can we simplify how we're setting the headers? Something like:

```ts
const headers = {
  ...instance.defaults.headers,
  Authorization: authenticate ? `Bearer ${contextState.user.accessToken}` : undefined,
  'Content-Type': body ? 'application/json' : undefined,
};
return instance.request({
  method,
  url,
  headers,
  data: body,
});
```

I'm pretty sure we can then remove both `if` blocks and this will still work if `body` is undefined, or `authenticate` is false, or neither or both.

ajrussellaudio (changes requested):

packages/account-management/src/utility/api-caller.ts
robmarch2: Unfortunately, this isn't possible. I'm getting an error with this when the headers are `undefined`:

```
service_1  | 16:31:07 0|server  | An error occurred in remote API call: TypeError [ERR_HTTP_INVALID_HEADER_VALUE]: Invalid value "undefined" for header "Content-Type"
```

ajrussellaudio (approved):

packages/account-management/src/utility/api-caller.ts
robmarch2: I'm going to leave it as is - I think conditionally setting the header is clearer than conditionally setting the header to a value or `undefined` then removing the `undefined`es

PR #72, opened 2021-01-22
Do not require user to login to access registration page

robmarch2:


packages/account-management/src/routes/index.ts
```diff
@@ -1,9 +1,12 @@
 import { RouteMiddleware } from '../types/application';
 
+const unAuthenticatedPages: string[] = ['/register'];
+
 export const indexPage: RouteMiddleware = (context) => {
   const bundle = context.routes.url('assets-bundles');
 
-  if (!context.isAuthenticated()) {
+  const isUnauthenticatedPage: boolean = !!context.req.url && unAuthenticatedPages.includes(context.req.url);
```
ajrussellaudio, line 8: Is there a situation where `req` wouldn't have a `url` here? Do we need to check that it exists?

robmarch2: In practice for us, probably not, but it's typed as optional with a comment that I think explains why:

```
        /**
         * Only valid for request obtained from http.Server.
         */
        url?: string;
```

robmarch2: Presumably there's a way of getting hold of the request that doesn't populate it

ajrussellaudio: Mental. Very philosophical: "What is a request, without a URL?"

We're good here though :+1:

ajrussellaudio (dismissed):

packages/account-management/src/routes/index.ts
```diff
@@ -5,7 +5,7 @@ const unAuthenticatedPages: string[] = ['/register', '/validated', '/error'];
 export const indexPage: RouteMiddleware = (context) => {
   const bundle = context.routes.url('assets-bundles');
 
-  const isUnauthenticatedPage: boolean = !!context.req.url && unAuthenticatedPages.includes(context.req.url);
+  const isUnauthenticatedPage: boolean = !!context.req.url && unAuthenticatedPages.includes(context.req.url.split('?')[0]);
```
ajrussellaudio, line 8: `context.request.URL.pathname` seems to give us what we're looking for here, and it isn't optional:

```suggestion
  const isUnauthenticatedPage: boolean = unAuthenticatedPages.includes(context.request.URL.pathname);
```

robmarch2: Good spot! That's much nicer

packages/account-management/src/routes/index.ts
```diff
@@ -5,8 +5,7 @@ const unAuthenticatedPages: string[] = ['/register', '/validated', '/error'];
 export const indexPage: RouteMiddleware = (context) => {
   const bundle = context.routes.url('assets-bundles');
 
-  const isUnauthenticatedPage: boolean = !!context.req.url && unAuthenticatedPages.includes(context.req.url.split('?')[0]);
-  if (!isUnauthenticatedPage && !context.isAuthenticated()) {
+  if (!unAuthenticatedPages.includes(context.request.URL.pathname) && !context.isAuthenticated()) {
```
ajrussellaudio, line 8: Sorry, this just came to me now that we've made the condition clearer. Would it be better to flip the logic?

```ts
  if (context.isAuthenticated() || unAuthenticatedPages.includes(context.request.URL.pathname)) {
    context.response.body = '<html />';
    return;
  }

  context.redirect('/login');
```

robmarch2: Yeah, I think that's a little better. Can get rid of the the `not unauthenticatedPage` double negative and make the redirect to login the default condition that way

packages/account-management/src/routes/index.ts
```diff
@@ -5,24 +5,24 @@ const unAuthenticatedPages: string[] = ['/register', '/validated', '/error'];
...
-        <script crossorigin src="https://unpkg.com/react@16/umd/react.development.js"></script>
-        <script crossorigin src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
-      </head>
-      <body>
-        <div id="root"></div>
-        <script type="application/javascript" src="${bundle}"></script>
-      </body>
-    </html>
-  `;
+  } else {
```
ajrussellaudio, line 25: If we're `return`ing above, we don't need the `else`

ajrussellaudio (changes requested):

ajrussellaudio (approved):

PR #73, opened 2021-01-22
Minor tidy ups

danielgrant:


packages/shared/auth0-client/src/auth0.ts
```diff
@@ -31,9 +31,8 @@ export function toAuth0Profile(auth0User: any): Auth0Profile {
   const userIdStr: string = auth0User.user_id;
   const name: string = auth0User.name;
   const email: string = auth0User.email;
-  const emailValidated: string = auth0User.email_verified;
   const creationDate: string = auth0User.created_at;
-  if (!isNonBlank(userIdStr) || !isNonBlank(name) || !isNonBlank(email) || !isNonBlank(emailValidated) || !isNonBlank(creationDate)) {
```
garyttierney, line 36: What do you think about introducing this in my next PR: https://express-validator.github.io/docs/?

The checks you're doing here can be expressed pretty much identically, but we can also get a bit of extra information about where the error happened.

garyttierney (approved):

PR #74, opened 2021-01-25
Setup Cypress

ajrussellaudio:
This PR:

- Installs and configures Cypress for local E2E testing
- Adds a couple of simple tests


packages/account-management/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
robmarch2, line 1: This doesn't look like it's used anywhere - can we get rid of it?

robmarch2 (approved):

stephenwf (approved):

ajrussellaudio:
Closing here, to replicate the work in https://github.com/wellcomecollection/wellcomecollection.org repo later

PR #80, opened 2021-02-02
Add endpoint to delete user records

garyttierney:


packages/shared/auth0-client/src/index.ts
```diff
@@ -31,6 +31,25 @@ export default class Auth0Client {
     this.clientSecret = clientSecret;
   }
 
+  async deleteUser(userId: number): Promise<APIResponse<boolean>> {
```
danielgrant, line 34: I have a concern over changes in structure between different functions. Ignoring the change in style, this function returns `boolean` to indicate success. By contrast, other functions with empty return types simply return `{}`, since it's the `ResponseStatus` which indicates success or failure to the caller.

In other words, the `boolean` returned by this is never used.

danielgrant (changes requested):
Can we make sure this is tested before merged? e.g. it doesn't include the modification to `auth0_client.buildkite` in the Terraform to add the required `delete:users` scope.

garyttierney:
Still needs tested first. @danielgrant do you have some in-flight Terraform changes being applied for the deletion request work? If so I'll get the other PR updated first and test when you're done.

danielgrant:
> Still needs tested first. @danielgrant do you have some in-flight Terraform changes being applied for the deletion request work? If so I'll get the other PR updated first and test when you're done.

Yeah good shout - if you `apply` this it'll undo my current working changes.

danielgrant:
@garyttierney this has been rebased on `main` on the Terraform applied - it should be good for testing.

danielgrant (approved):

PR #92, opened 2021-02-17
WS-127 Add template for user email notifying of delete request

ajrussellaudio:


packages/apps/api/src/utils/email.ts
```diff
@@ -39,7 +39,8 @@ export default class EmailClient {
     const subject: string = await this.engine.renderFile('delete-request_user_subject', {});
     const body: string = await this.engine.renderFile('delete-request_user_body', {
       firstName: auth0Profile.firstName,
-      lastName: auth0Profile.lastName
+      lastName: auth0Profile.lastName,
+      supportUrl: 'https://wellcome.org/about-us/contact-us'
```
ajrussellaudio, line 43: This is not the right URL.

The real URL is stored in a SSM value, accessed in Terraform: `aws_ssm_parameter.auth0_support_url.value`

I'm assuming that SSM values cannot be accessed here. Is hard-coding this value the right approach? What should the correct URL be?

danielgrant (approved):

PR #100, opened 2021-03-11
Enabling the Auth0 Welcome Email

danielgrant:


infra/scoped/parameters.tf
```diff
@@ -233,6 +233,23 @@ resource "aws_ssm_parameter" "auth0_reset_email_url_ttl" {
   )
 }
 
+resource "aws_ssm_parameter" "auth0_welcome_email_subject" {
+  name  = "identity-auth0_welcome_email_subject-${terraform.workspace}"
```
garyttierney, line 237: Does this need to use the prefix you generated in the other PR?

danielgrant: Nah, this isn't a user-facing resource. That qualifier is only for the resources that a user can actually see.

garyttierney (approved):

PR #109, opened 2021-04-09
Move some terraform around to be more predictable

kenoir:
This change creates a `locals.tf` and a `terraform.tf` as it is usual for Wellcome projects to split out terraform in that way.

The intention is to make the project a bit easier to navigate when making the required changes for requesting.


danielgrant (dismissed):

danielgrant:
@kenoir could you fix the pre-commit linting issue?

danielgrant (approved):

PR #111, opened 2021-04-09
Assume identity-developer role in provider/backend

kenoir:
This change uses `assume_role` to configure provider & backend with the correct permissions. 

Bringing this slightly more in-line with other `wellcomecollection` projects.

Also adds `.terraform.lock.hcl` to source control in line with recommendation: https://www.terraform.io/docs/language/dependency-lock.html

### DIFF
--- a/infra/scoped/.terraform.lock.hcl
+++ b/infra/scoped/.terraform.lock.hcl
@@ -1,0 +1,111 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/alexkappa/auth0" {
+  version     = "0.16.1"
+  constraints = "~> 0.16.0"
+  hashes = [
+    "h1:Yuan3uU/eISWfu9f7JMG8WackUOMnrrh3LFywdxqMek=",
+    "zh:173017f420275c617881602a35fda7beb6e097d6d32fbdbb035046b4a90a4136",
+    "zh:2422e7e41fea5e9578e9fd5c71376ab48328141b90ffd7e2c0823d581b309563",
+    "zh:433c7cac260c7882d4f889d097907ea779fb7d890366109ca9c3d18ad056aac9",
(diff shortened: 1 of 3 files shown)


kenoir:
Failed at https://buildkite.com/wellcomecollection/identity/builds/726#75aab50d-62b3-4ac2-aaa7-77c321d106a9/122-133

```
Error: error configuring S3 Backend: IAM Role (arn:aws:iam::770700576653:role/identity-developer) cannot be assumed.
--
  |  
  | There are a number of possible causes of this - the most common are:
  | * The credentials used in order to assume the role are invalid
  | * The credentials do not have appropriate permission to assume the role
  | * The role ARN is not valid
  |  
```

`terraform` runs in CI here unlike our other projects - we need to accommodate continuing to run tf.

alexwlchan:
This confused me initially, I couldn't find my way into the identity account in the AWS console.

alexwlchan (approved):

kenoir:
Ping @danielgrant. I think this has the potential to change dev workflow from your side, so waiting on your review to merge this.

danielgrant:
> Ping @danielgrant. I think this has the potential to change dev workflow from your side, so waiting on your review to merge this.

Yes, this does cause problems for local development.

I'm not sure why this change is being introduced, we already explicitly set the role we want to use during CI [here](https://github.com/wellcomecollection/identity/blob/main/.buildkite/pipeline.deploy.yml#L27). This change just repeats that setting explicitly in the Terraform but it's not clear why you want to do that?

kenoir:
@danielgrant the way we work in other Wellcome Collection repos, we set the role in the provider so that we do not have to hand it to the project in another way. We would expect to be able to run this terraform locally as easily as possible.

A key difference here is probably that this project runs terraform in CI - and we have avoided doing that elsewhere.

kenoir:
Closing as not essential at the moment.

PR #125, opened 2021-05-06
Log mostly at the debug lebel; use template strings

jamieparkinson:


packages/apps/api-authorizer/src/index.ts
```diff
@@ -89,18 +81,11 @@ export async function lambdaHandler(
...
-        ' ' +
-        event.resource +
-        '] with path parameters [' +
-        event.pathParameters +
-        ']'
+    console.debug(
+      `Access token [${accessToken}] for user [${JSON.stringify(
+        auth0UserInfo
+      )}]` +
+        `cannot operate on [${event.httpMethod} ${event.resource}] with path parameters [${event.pathParameters}]`
```
alexwlchan, line 88: Probably not one to fix now, but we should look at exactly what's in that `auth0UserInfo` variable before we go logging it everywhere.

packages/apps/api/src/utils/email.ts
```diff
@@ -113,18 +113,13 @@ export default class EmailClient {
...
       })
       .catch((error) => {
-        console.log(
-          'An error occurred sending SMTP email [' +
-            smtpParams +
-            ']: [' +
-            error +
-            ']'
+        console.error(
+          `An error occurred sending SMTP email [${smtpParams}]`,
```
alexwlchan, line 121: Ditto `smtpParams` and possible user email addresses.

packages/shared/auth0-client/src/auth0.ts
```diff
@@ -84,12 +84,11 @@ function toAuth0Profiles(users: any[]): Auth0Profile[] {
...
     } catch (e) {
-      console.log(
-        'An error occurred converting Auth0 user [' +
-          JSON.stringify(user) +
-          '] to Auth0Profile: [' +
-          e +
-          ']'
+      console.error(
+        `An error occurred converting Auth0 user [${JSON.stringify(
+          user
```
alexwlchan, line 89: Ditto.

alexwlchan (approved):

packages/apps/api-authorizer/src/index.ts
jamieparkinson: shouldn't be an issue at the debug level, but yes, should be done especially for the other cases

kenoir: Definitely PII: https://github.com/wellcomecollection/identity/blob/45c9badfc2aa8ba7197530b1a64193836a0056fb/packages/shared/auth0-client/src/auth0.ts#L151

kenoir (commented):
Good points about logging PII raised here.

packages/apps/api-authorizer/src/index.ts
alexwlchan: > shouldn't be an issue at the debug level

Not sure I agree. I don't love the idea that changing the log level suddenly starts putting PII into our logging cluster.

alexwlchan:
I've created a ticket to track the PII discussions: https://github.com/wellcomecollection/platform/issues/5166

packages/apps/api-authorizer/src/index.ts
jamieparkinson: True (although I got the [idea](https://wellcome.slack.com/archives/CUA669WHH/p1618396869015100?thread_ts=1618395149.009100&cid=CUA669WHH) from you 😬) - I will return to this in my next PR

alexwlchan: The person who wrote that message looks very shifty. I wouldn't trust anything they say! 😉

jamieparkinson: Encouraging accidental leaking of PII is the kind of thing a spy might do

PR #129, opened 2021-05-12
Add tests for api-authorizer

jamieparkinson:
Thought this was a good start for familiarising myself with the codebase; I split up the main file a little in order to make dependency injection easier without causing side effects etc. It's been a long time since I had to do JS testing and I have not missed it!


packages/apps/api-authorizer/src/index.ts
```diff
@@ -1,277 +1,21 @@
-import Auth0Client, { Auth0UserInfo } from '@weco/auth0-client';
```
kenoir, line 1: Does this file need to continue to exist of should we fold it into handler.ts?

kenoir (approved):

kenoir:
Nice work 👏

packages/apps/api-authorizer/src/index.ts
jamieparkinson: It does - because the redis/auth0 services live at the module scope, importing it will initialise them (which throws an error in tests). Keeping it like this to export the "real" version of the handler links up with [here](https://github.com/wellcomecollection/identity/blob/ca609e1c475826889b578649972db8bb79833a53/infra/scoped/lambda.tf#L5)

PR #130, opened 2021-05-13
URL encode pin when sending to the /patrons/validate endpoint

kenoir:
If passwords are not urlencoded when sent to the /patrons/validate endpoint they will not be correctly validated. This change encodes those passwords.

This change ensures that users who have [url-encodable characters](https://en.wikipedia.org/wiki/Percent-encoding) can be transparently migrated from Wellcome Library to Wellcome Collection.

### DIFF
--- a/packages/apps/auth0-actions/src/login.js
+++ b/packages/apps/auth0-actions/src/login.js
@@ -36,7 +36,7 @@ function login(email, password, callback) {
           '/v6/patrons/validate',
           {
             barcode: barcode,
-            pin: pin,
+            pin: encodeURI(pin),
           },
           {
             validateStatus: (status) => status === 204,
--- a/packages/apps/auth0-actions/src/login.js
+++ b/packages/apps/auth0-actions/src/login.js
@@ -29,6 +29,9 @@ function login(email, password, callback) {
       }
     });
 
+  // The /v6/patrons/validate expects passwords encoded using the encodeURIComponent scheme.
+  // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI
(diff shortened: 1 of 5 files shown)


kenoir:
This PR needs to add some tests for the auth0-actions before merging.

packages/apps/auth0-actions/src/login.js
```diff
@@ -36,7 +36,7 @@ function login(email, password, callback) {
           '/v6/patrons/validate',
           {
             barcode: barcode,
-            pin: pin,
+            pin: encodeURI(pin),
```
jamieparkinson, line 39: Will this work? See "not escaped" list here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI#description

jamieparkinson: `encodeURIComponent` will handle more, but not the `!` - what encoding did you find worked for that? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#description

danielgrant: FWIW I didn't have a problem with `!` when I was testing this directly against the Sierra API.

kenoir: > FWIW I didn't have a problem with ! when I was testing this directly against the Sierra API.
Indeed testing this in isolation (in a different password from what I was using before) it does seem to be fine with ! not encoded.

kenoir: Testing makes it look like we need encodeURIComponent here.

jamieparkinson (approved):
good comment

PR #132, opened 2021-05-14
Add requests app ECS & LB

jamieparkinson:
This is the first part of the requests app infra, most of the diff is [this module](https://github.com/wellcomecollection/catalogue-api/tree/main/terraform/modules/service) copied and pasted from the catalogue-api repo.


kenoir:
Don't forget to ignore the deployment label tag as in: https://github.com/wellcomecollection/catalogue-api/blob/main/terraform/catalogue_api/provider.tf#L8

infra/modules/ecs_service/variables.tf
```diff
@@ -0,0 +1,68 @@
...
+
+variable "secrets" {
+  type    = map(string)
+  default = {}
+}
+
+variable "container_port" {
+  type    = number
+  default = 9001
+}
```
alexwlchan, line 22: Nit: do we want to set a default here? Feels like there's an easy mistake where you forget to set it and wonder why you don't see anything.

infra/scoped/ecs_services.tf
```diff
@@ -0,0 +1,44 @@
...
+module "requests" {
+  source = "../modules/ecs_service"
+
+  service_name            = "requests-${terraform.workspace}"
+  deployment_service_name = "requests"
+
+  container_image        = "${local.requests_repository}:env.${terraform.workspace}"
+  container_port         = local.requests_container_port
+  deployment_service_env = terraform.workspace
+  desired_task_count     = 1
```
alexwlchan, line 14: Why only 1?

infra/scoped/ecs_services.tf
```diff
@@ -0,0 +1,44 @@
...
+
+  load_balancer_arn           = aws_lb.identity_api.arn
+  load_balancer_listener_port = local.requests_lb_port
+
+  cluster_arn = aws_ecs_cluster.identity.arn
+  vpc_id      = aws_vpc.main.id
+  subnets     = local.private_subnets
+
+  security_group_ids = [
+    aws_security_group.local.id,
```
alexwlchan, line 40: What's the local security group? Do we need it?

alexwlchan (dismissed):

infra/scoped/ecs_services.tf
jamieparkinson: It's what we have called `ingress` elsewhere - local in the sense that it allows ingress from within the VPC

infra/scoped/ecs_services.tf
jamieparkinson: true, I'll make this the same as how we do tasks and AZs in the catalogue API

jamieparkinson:
(these tests are failing at the moment because I've bumped a module version pre-emptively, cf https://github.com/wellcomecollection/terraform-aws-ecs-service/pull/34)

infra/modules/ecs_service/main.tf
```diff
@@ -0,0 +1,80 @@
...
+  source    = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/secrets?ref=v3.6.2"
+  secrets   = var.secrets
+  role_name = module.task_definition.task_execution_role_name
+}
+
+module "task_definition" {
+  source = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/task_definition?ref=v3.6.2"
+
+  cpu    = 1024
+  memory = 2048
```
alexwlchan, line 43: Minor: worth making these variables?

infra/modules/ecs_service/target_group.tf
```diff
@@ -0,0 +1,32 @@
...
+
+  target_type = "ip"
+
+  protocol = "TCP"
+  port     = module.nginx_container.container_port
+  vpc_id   = var.vpc_id
+
+  # The default deregistration delay is 5 minutes, which means that ECS
+  # takes around 5–7 mins to fully drain connections to and deregister
+  # the old task in the course of its blue/green. deployment of an
```
alexwlchan, line 14: nit:

```suggestion
  # the old task in the course of its blue/green deployment of an
```

alexwlchan (approved):

Issue #133, opened 2021-05-14
Add authenticated smoke test for Identity API


kenoir:
We should have smoke & e2e tests as with the catalogue API to exercise key functionality when deploying to stage & prod.

This will require authenticating a test user and performing core operations on the API.

This is in order to avoid deploying broken things!

kenoir:
Closing as done.

PR #137, opened 2021-05-18
Add simple smoke tests for identity API

kenoir:
In order to be comfortable regularly deploying the identity APIs we want to be sure that a deployment has been successful. This change adds the framework for (and an initial example of) simple smoke tests to run against deployed services to ensure they are operating as expected.

When run this test will make an authenticated request against the /users/me endpoint and check that a 200 OK is returned.

This change uses a test client application set up in Auth0 to allow the [Resource Owner Password Flow](https://auth0.com/docs/flows/resource-owner-password-flow).

We are using callbacks here because async/await & Promises are not available in artillery.io processors.

Remaining to do:
- [x] Provision auth0 client app in terraform for stage & prod
- [x] Hook into buildkite deployment pipeline
- [x] Put test user credentials in secrets manager


kenoir:
For https://github.com/wellcomecollection/identity/issues/133

infra/scoped/locals.tf
```diff
@@ -6,6 +6,10 @@ locals {
     "ManagedBy"   = var.tag_managed_by
   }
 
+  stage_test_client_ids = [
+    # Developer client ids can be added here
```
kenoir, line 10: I think we should probably define the developer clients in TF - better that than either blatting them in an apply or excluding them.

smoke_tests/processorFunctions.js
```diff
@@ -0,0 +1,104 @@
+const AWS = require('aws-sdk');
```
jamieparkinson, line 1: Can you link to the docs & maybe add some comments in this file? It is quite hard to tell what it does otherwise, as how the Artillery scripting stuff works is not very obvious

smoke_tests/processorFunctions.js
```diff
@@ -0,0 +1,104 @@
...
+
+const prodSecrets = {
+  auth0Url: 'https://wellcomecollection.eu.auth0.com/oauth/token',
+  apiKeySecretId: 'identity/prod/account_management_system/api_key',
+  credentialsSecretId: 'identity/prod/smoke_test/credentials',
+};
+
+const secretsManager = new AWS.SecretsManager();
+
+let apiKey;
```
jamieparkinson, line 22: When does this get set?

smoke_tests/processorFunctions.js
```diff
@@ -0,0 +1,104 @@
...
+
+function addCredentials(requestParams, context, ee, next) {
+  let isStage = context['vars']['$environment'] === 'stage';
+  const secrets = isStage ? stageSecrets : prodSecrets;
+
+  if (apiKey && accessToken) {
+    requestParams.headers['x-api-key'] = apiKey;
+    requestParams.headers['Authorization'] = accessToken;
+    next();
+  } else {
```
jamieparkinson, line 91: Is this branch ever taken?

jamieparkinson (approved):

smoke_tests/processorFunctions.js
kenoir: This is some vestigial caching stuff for when we make multiple requests - I should make it work again.

smoke_tests/processorFunctions.js
kenoir: Same as above.

PR #141, opened 2021-05-19
Upgrade AWS provider & use default tags

jamieparkinson:
Saw this when I was about to do the VPC stuff and thought I'd make the change


alexwlchan (approved):
You've buried the lede here – default tags are perfect for spreading the Terraform configuration URL tags! 😍

PR #145, opened 2021-05-24
Add terraform runner script

kenoir:
In order to run terraform without having to copy paste credentials from Auth0, this change pulls the buildkite credentials from secrets manager and applies those to the running environment.

This allows terraform to be run in CI or locally and simplifies the process of running terraform. Using this change removes the need to have a client for each user and to maintain that either in or out of terraform. 

This change does remove one part of the audit trail to see who (i.e. which client) has made changes in Auth0, but at present that is of dubious value.

**Note:** The values required to be set in SM/SSM have been manually set for _stage_ only currently, they will need to be added for _prod_ before merging. We could set those values from (buildkite client id/secret) if there is appetite to do so.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,12 +21,6 @@ repos:
       exclude: ^infra/scoped/task-definitions
     - id: check-yaml
 
-- repo: https://github.com/detailyang/pre-commit-shell.git
-  rev: v1.0.6
-  hooks:
-    - id: shell-lint
-      args: [--format=json]
-
 - repo: https://github.com/bemeurer/beautysh.git
   rev: 6.0.1
   hooks:
--- a/infra/scoped/backend.tf
+++ b/infra/scoped/backend.tf
@@ -1,6 +1,4 @@
 terraform {
-  required_version = "= 0.14.2" # Pin to a specific version to avoid accidental upgrading of the statefile
-
   backend "s3" {
     role_arn = "arn:aws:iam::770700576653:role/identity-developer"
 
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -21,12 +21,6 @@ repos:
       exclude: ^infra/scoped/task-definitions
     - id: check-yaml
 
-- repo: https://github.com/detailyang/pre-commit-shell.git
```
kenoir, line 24: This check is removed as it failed the pre-commit check and did not provide useful output (the script passes when sent through https://www.shellcheck.net/).

infra/scoped/backend.tf
```diff
@@ -1,6 +1,4 @@
 terraform {
-  required_version = "= 0.14.2" # Pin to a specific version to avoid accidental upgrading of the statefile
```
kenoir, line 2: Unpin the terraform version as we don't have a mechanism so we do not have to explicitly care about which version we are using in the runner script. 

We do not pin terraform in any other project, so this is more familiar to Wellcome developers, although it _will_ cause the TF version to roll forward with all the associated issues.

infra/scoped/run_terraform.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+TERRAFORM_WORKSPACE=$(terraform workspace show)
+
+AWS_CLI_PROFILE="identity-terraform"
```
kenoir, line 5: Instead of relying on a user to set the correct profile as in https://github.com/wellcomecollection/platform-infrastructure/blob/main/critical/run_terraform.sh, we create a custom profile and assign it a known role.

.pre-commit-config.yaml
jamieparkinson: fwiw the way this pre-commit tool works is it fails if it makes any changes, so a second attempt works. I had a bit of confusion with that - still think it's worth removing this specific check

jamieparkinson (approved):

Issue #136, opened 2021-05-17
Custom authorizer does not distinguish between unauthorized and forbidden requests


jamieparkinson:
Putting this here while I wrangle APIGW boilerplate, I had mistakenly thought there would be 401s but there aren't currently

https://stackoverflow.com/questions/51753487/custom-response-lambda-authorizer-for-401/55608511

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html#api-gateway-lambda-authorizer-lambda-function-create

PR #150, opened 2021-05-26
Reject unauthorized requests with a 401, not a 403

jamieparkinson:
This is very pedantic but there we are - resolves https://github.com/wellcomecollection/identity/issues/136

### DIFF
--- a/infra/scoped/api-documentation.tf
+++ b/infra/scoped/api-documentation.tf
@@ -173,6 +173,20 @@ resource "aws_api_gateway_documentation_part" "users_get_400" {
   }
 }
 
+# 401 Unauthorized
+
+resource "aws_api_gateway_documentation_part" "users_get_401" {
+  rest_api_id = aws_api_gateway_rest_api.identity.id
+  properties  = file("${path.module}/api-documentation/users/get-401.json")
+
+  location {
+    type        = "RESPONSE"
+    path        = aws_api_gateway_resource.users.path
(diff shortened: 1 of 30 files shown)


infra/scoped/api-documentation.tf
```diff
@@ -1102,6 +1200,20 @@ resource "aws_api_gateway_documentation_part" "users_userid_lock_put_200" {
   }
 }
 
+# 401 Unauthorized
+
+resource "aws_api_gateway_documentation_part" "users_userid_lock_put_401" {
+  rest_api_id = aws_api_gateway_rest_api.identity.id
+  properties  = file("${path.module}/api-documentation/users/:user_id/lock/put-401.json")
```
jamesgorrie, line 1207: Worth having these `properties` as a list in  a `local` and looping through em?

e.g.
```tf
locals {
  put401paths = [ "https://github.com/wellcomecollection/catalogue-pipeline/pull/1639",... ]
}
```

jamesgorrie (approved):
Nice just wonder if there's a nicer way to have config that provides certain TF response resources.

infra/scoped/api-documentation.tf
jamieparkinson: Yeah, I think there is a larger (in fact, really quite large) piece of work to reduce the boilerplate in this terraform - it's all very flat at the moment which is fine but I think we could write some modules which work per-route, rather than per-terraform-resource.

PR #151, opened 2021-06-16
Add smoke test for /users/:id:/item-requests endpoint

kenoir:
Add a smoke test for the `/users/me/item-requests` endpoint endpoint to increase deployment reliability.


kenoir:
This test fails when run locally now - I am investigating.

jamieparkinson:
Sometimes can be very slow because of Sierra...

PR #153, opened 2021-06-17
Update README to include some docs about build

kenoir:
Update the docs to describe the build process a bit better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,9 @@ Switched to workspace "stage".
 
 This project builds in Buildkite where there are pipelines to cover build and deployment through stage and production environments.
 
-- [https://buildkite.com/wellcomecollection/identity](Identity): Test and build artifacts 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,3 +23,28 @@ You will need to switch workspace to plan/apply in a particular environment:
 > terraform workspace select stage
 Switched to workspace "stage".
 ```
+
+### Continuous Integration
+
+This project builds in Buildkite where there are 3 pipelines to cover build and deployment through stage and production environments.
```
alexwlchan, line 29: Three pipelines but only two environments? This sentence makes me think the two numbers are related, but maybe they're not?

kenoir: There's one for build and two for deployment. Probably worth clarifying here.

alexwlchan (approved):

PR #154, opened 2021-06-28
Use Sierra auth endpoint

jamieparkinson:
This also includes a bit of a refactor of the login script and removes the duplicate (unused) function from the Sierra client; I'm going to spend a little time now looking at refactoring the other scripts and maybe making another attempt at Typescripting them...


jamieparkinson:
Going to do this in sierra-client after converting hooks to TS

PR #159, opened 2021-07-01
Return a more informative error from the login scripts

jamieparkinson:
Auth0 gives us some error classes to communicate different error causes - this gives us control over the microcopy, and fixes the bug that incorrect credentials on a first login returned a "something went wrong" error.

The copy used here is OK for now - see [Slack discussion](https://wellcome.slack.com/archives/C294K7D5M/p1625133599118500)

### DIFF
--- a/packages/apps/auth0-actions/src/login.ts
+++ b/packages/apps/auth0-actions/src/login.ts
@@ -9,6 +9,9 @@ declare const configuration: {
   CLIENT_SECRET: string;
 };
 
+const invalidCredentialsMessage =
+  "We don't recognise the email and/or password you entered. Please check your entry and try again";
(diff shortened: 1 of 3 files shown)


packages/apps/auth0-actions/src/login.ts
```diff
@@ -9,6 +9,9 @@ declare const configuration: {
   CLIENT_SECRET: string;
 };
 
+const invalidCredentialsMessage =
+  "We don't recognise the email and/or password you entered. Please check your entry and try again";
```
alexwlchan, line 13: nit: Does this need a full stop?

alexwlchan (approved):

PR #160, opened 2021-07-01
Add tests for Auth0 actions

jamieparkinson:
Does what it says on the tin, I also tightened up some of the types a bit.

### DIFF
--- a/packages/apps/auth0-actions/jest.setup.js
+++ b/packages/apps/auth0-actions/jest.setup.js
@@ -1,0 +1,16 @@
+// This replicates (some of) the globals that are available in
+// an Auth0 script environment
+
+global.configuration = {
+  API_ROOT: 'test_api_host',
+  CLIENT_KEY: 'test_client_key',
+  CLIENT_SECRET: 'test_client_secret',
+};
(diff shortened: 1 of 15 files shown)


packages/apps/auth0-actions/jest.setup.js
```diff
@@ -0,0 +1,16 @@
+// This replicates (some of) the globals that are available in
+// an Auth0 script environment
+
+global.configuration = {
+  API_ROOT: 'test_api_host',
+  CLIENT_KEY: 'test_client_key',
+  CLIENT_SECRET: 'test_client_secret',
+};
+
+class WrongUsernameOrPasswordError extends Error {
```
kenoir, line 10: This is duplicated from `databaseScriptGlobals.d.ts`. Is it not possible to share the definition?

jamieparkinson: Unfortunately not really - that one is a type definition, this is a class definition (see that the one in the `.d.ts` file doesn't actually implement the constructor)

kenoir (approved):

PR #175, opened 2021-08-31
Fix 500 for unauthenticated "me" calls

jamieparkinson:
This makes the identity API use the same error handling logic everywhere (hence the large diff). 

There are no tests for this app, so those are coming up in a separate PR to make sure I haven't broken anything.

The part that fixes the bug is https://github.com/wellcomecollection/identity/pull/175/files#diff-e7bd22e600a28efd6644ba0a4f4d5c42ecd9c154360e1e0aa162df38ffbca58dR641-R669

### DIFF
--- a/packages/apps/api/src/handlers/errorHandler.ts
+++ b/packages/apps/api/src/handlers/errorHandler.ts
@@ -13,7 +13,5 @@ export const errorHandler = (
   }
 
   const status = err instanceof HttpError ? err.status : 500;
-  const message = err.message ?? 'Something went wrong';
-
-  res.status(status).json(toMessage(message));
+  res.status(status).json(err.message ? toMessage(err.message) : undefined);
 };
--- a/packages/apps/api/src/handlers/errorHandler.ts
+++ b/packages/apps/api/src/handlers/errorHandler.ts
@@ -1,0 +1,19 @@
+import { NextFunction, Request, Response } from 'express';
+import { HttpError } from '../models/HttpError';
+import { toMessage } from '../models/common';
+
+export const errorHandler = (
+  err: Error,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (res.headersSent) {
+    return next(err);
+  }
+
+  const status = err instanceof HttpError ? err.status : 500;
+  const message = err.message ?? 'Something went wrong';
+
+  res.status(status).json(toMessage(message));
+};
--- a/packages/apps/api/src/handlers/user.ts
+++ b/packages/apps/api/src/handlers/user.ts
@@ -27,7 +27,7 @@ export async function validatePassword(
   if (!isNonBlank(password)) {
     throw new HttpError({
       status: 400,
-      message: 'All fields must be provided and non-blank',
+      message: 'A password must be provided and non-blank',
     });
   }
 
@@ -93,7 +93,7 @@ export async function createUser(
   ) {
     throw new HttpError({
       status: 400,
-      message: 'All fields must be provided and non-blank',
+      message: 'A password must be provided and non-blank',
     });
   }
 
@@ -106,9 +106,9 @@ export async function createUser(
       message: `Patron record with email [${email}] already exists`,
     });
   } else if (sierraGet.status !== ResponseStatus.NotFound) {
+    console.error('Unexpected error from Sierra client', sierraGet.message);
     throw new HttpError({
       status: 500,
-      message: 'Something went wrong: ' + sierraGet.message,
     });
   }
 
@@ -121,9 +121,9 @@ export async function createUser(
       message: `Auth0 user with email [${email}] already exists`,
     });
   } else if (auth0Get.status !== ResponseStatus.NotFound) {
+    console.error('Unexpected error from Auth0 client', auth0Get.message);
     throw new HttpError({
       status: 500,
-      message: 'Something went wrong: ' + auth0Get.message,
     });
   }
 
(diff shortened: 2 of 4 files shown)


packages/apps/api/src/models/HttpError.ts
```diff
@@ -0,0 +1,37 @@
...
+        case ResponseStatus.MalformedRequest:
+          return 400;
+        case ResponseStatus.PasswordTooWeak:
+          return 422;
+        case ResponseStatus.QueryTimeout:
+          return 503;
+        case ResponseStatus.RateLimited:
+          return 429;
+        case ResponseStatus.UnknownError:
+          return 500;
```
alexwlchan, line 33: Do we want a fallthrough case for "any other status" that logs a warning and maps to 500?

jamieparkinson: Nope, these are checked for exhaustiveness by the type system - any uncaught errors will be returned as 500s anyway

alexwlchan: niiiiiiiice

packages/apps/api/src/handlers/errorHandler.ts
```diff
@@ -0,0 +1,19 @@
...
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (res.headersSent) {
+    return next(err);
+  }
+
+  const status = err instanceof HttpError ? err.status : 500;
+  const message = err.message ?? 'Something went wrong';
```
alexwlchan, line 16: Very minor: our approach elsewhere is to omit the message if there's nothing useful to say.

packages/apps/api/src/handlers/user.ts
```diff
@@ -113,42 +91,40 @@ export async function createUser(
...
+  if (sierraGet.status === ResponseStatus.Success) {
+    throw new HttpError({
+      status: 409,
+      message: `Patron record with email [${email}] already exists`,
+    });
+  } else if (sierraGet.status !== ResponseStatus.NotFound) {
+    throw new HttpError({
+      status: 500,
+      message: 'Something went wrong: ' + sierraGet.message,
+    });
```
alexwlchan, line 112: Possibly not for addressing here, but elsewhere we don't include any context in 500 errors – we return a generic error and log the message.

packages/apps/api/src/handlers/user.ts
```diff
@@ -264,63 +214,47 @@ export async function updateUser(
...
-        response.status(500).json(toMessage(auth0EmailGet.message));
+    if (auth0EmailGet.status === ResponseStatus.Success) {
+      if (auth0EmailGet.result.userId !== userId.toString()) {
+        throw new HttpError({
+          status: 409,
+          message: `Auth0 user with email [${fields.email.value}] already exists`,
+        });
       }
-      return;
+    } else if (auth0EmailGet.status !== ResponseStatus.NotFound) {
```
alexwlchan, line 236: Where do we handle the case where the auth0EmailGet returns a 404?

packages/apps/api/src/handlers/user.ts
```diff
@@ -371,44 +289,26 @@ export async function changePassword(
 
   const password: string = request.body.password;
   if (!isNonBlank(password)) {
-    response
-      .status(400)
-      .json(toMessage('All fields must be provided and non-blank'));
-    return;
+    throw new HttpError({
+      status: 400,
+      message: 'All fields must be provided and non-blank',
```
alexwlchan, line 294: Minor: should we return a more specific error message here? We know the password field is non-blank, are there any other fields we're expecting?

alexwlchan: cf the error message further down this diff

packages/apps/api/src/handlers/user.ts
```diff
@@ -791,63 +611,61 @@ export async function deleteUser(
...
   return request.apiGateway?.event.requestContext.identity.sourceIp;
 }
 
 function getTargetUserId(request: Request): number {
   if (request.params.user_id === 'me') {
     if (userIsAdmin(request)) {
-      throw new Error('Administrator users cannot operate on themselves');
+      throw new HttpError({
+        status: 404,
+        message: 'Administrator users cannot operate on themselves',
```
alexwlchan, line 646: Would a 403 Forbidden be more appropriate than a 404 here?

packages/apps/api/src/handlers/user.ts
```diff
@@ -791,63 +611,61 @@ export async function deleteUser(
...
       request.apiGateway?.event.requestContext.authorizer?.callerId
     );
     if (isNaN(callerId)) {
-      throw new Error(
-        `Caller attempted to operate on themself, but has an invalid user ID: [${callerId}]`
-      );
+      throw new HttpError({
+        status: 401,
+        message: `Caller attempted to operate on themselves, but has an invalid user ID: [${callerId}]`,
+      });
```
alexwlchan, line 657: Invalid user ID feels more like a 400? Doesn't seem like providing authorisation would make this okay (same below).

alexwlchan (approved):
Bulk of the diff looks good, just a couple of comments inline.

packages/apps/api/src/handlers/user.ts
jamieparkinson: We want it to return a 404!

packages/apps/api/src/handlers/user.ts
jamieparkinson: Hmm yeah probably

packages/apps/api/src/handlers/user.ts
jamieparkinson: invalid is probably the wrong term - I'll tweak this

packages/apps/api/src/handlers/user.ts
jamieparkinson: I was just replicating existing behaviour here, but yeah I'll tweak this copy

PR #176, opened 2021-09-03
Catch errors thrown in async route handlers

jamieparkinson:
Introduced this issue in https://github.com/wellcomecollection/identity/pull/175, found it in https://github.com/wellcomecollection/identity/pull/177

### DIFF
--- a/packages/apps/api/package.json
+++ b/packages/apps/api/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "aws-serverless-express": "^3.3.8",
     "express": "^4.17.1",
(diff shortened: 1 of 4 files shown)


alexwlchan (approved):
It looks reasonable to me, but I'm decidedly non-familiar with JS stuff.

jamieparkinson:
The diff is a bit confusing - it makes the route handlers partially applied (for readability), injecting the dependencies on the first application and taking requests/responses thereafter, and wraps them in a utility library which makes sure errors are caught by the middleware

PR #178, opened 2021-09-03
Don't tell users who are already in Sierra to verify their emails

jamieparkinson:
Does what it says on the tin

### DIFF
--- a/packages/apps/auth0-actions/src/login.ts
+++ b/packages/apps/auth0-actions/src/login.ts
@@ -37,7 +37,16 @@ async function login(email: string, password: string): Promise<User> {
     throw new WrongUsernameOrPasswordError(email, invalidCredentialsMessage);
   }
 
-  return patronRecordToUser(patronRecord);
+  const user = patronRecordToUser(patronRecord);
+  return {
+    ...user,
+    // This script executes when a user who is not found in the Auth0 database attempts to log in.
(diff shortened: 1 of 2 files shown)


packages/apps/auth0-actions/src/login.ts
```diff
@@ -37,7 +37,16 @@ async function login(email: string, password: string): Promise<User> {
     throw new WrongUsernameOrPasswordError(email, invalidCredentialsMessage);
   }
 
-  return patronRecordToUser(patronRecord);
+  const user = patronRecordToUser(patronRecord);
+  return {
+    ...user,
```
kenoir, line 42: I think it's possible for accounts to exist in Sierra with unvalidated e-mail addresses, though practically we are unlikely to encounter that situation or for it to be an issue.

kenoir (approved):

packages/apps/auth0-actions/src/login.ts
jamieparkinson: Do the patron records tell us whether they're validated? If so easy enough to account for that

kenoir: I do not know. There is nothing obvious on the [patron record](https://sandbox.iii.com/iii/sierra-api/swagger/index.html#!/patrons/Get_a_patron_by_record_ID_get_18) that indicates that. @pollecuttn may know how that's distinguished.

I wouldn't let that block  merging this though.

PR #183, opened 2021-09-09
Terraform e-mail setup more extensively

kenoir:
In order that future engineers can understand the SMTP configuration for sending e-mails (e.g. when we want to send password reset e-mails, or notify library support of requested account deletions) we should fully describe the setup in this repository.

This change will help future engineers make changes or support issues where we need to update e-mail credentials.

This includes:
- Terraform our mailtrap.io configuration
- Hold SES configuration in Secrets Manager
- Not requiring manual setup of secrets in AWS secrets manager when credentials change


infra/scoped/iam.tf
```diff
@@ -60,38 +60,6 @@ resource "aws_iam_role_policy_attachment" "AWSLambdaVPCAccessExecutionRole" {
   policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole"
 }
 
-# SES
-
-resource "aws_iam_user" "auth0_email" {
```
kenoir, line 65: This has moved to the static stack - we don't need a user per env as we only send SES e-mails in production.

jamieparkinson (approved):
🎉

infra/scoped/lambda.tf
```diff
@@ -96,10 +96,10 @@ resource "aws_lambda_function" "api" {
...
       AUTH0_CLIENT_SECRET  = auth0_client.api_gateway_identity.client_secret,
       API_ALLOWED_ORIGINS  = local.identity_v1_docs_endpoint,
-      EMAIL_SMTP_HOSTNAME  = aws_ssm_parameter.email_smtp_hostname.value,
-      EMAIL_SMTP_PORT      = aws_ssm_parameter.email_smtp_port.value,
-      EMAIL_SMTP_USERNAME  = aws_ssm_parameter.email_smtp_username.value,
-      EMAIL_SMTP_PASSWORD  = data.aws_secretsmanager_secret_version.email_smtp_password_secret_version.secret_string,
+      EMAIL_SMTP_HOSTNAME  = local.email_credentials["smtp_hostname"],
+      EMAIL_SMTP_PORT      = local.email_credentials["smtp_port"],
+      EMAIL_SMTP_USERNAME  = local.email_credentials["smtp_username"],
+      EMAIL_SMTP_PASSWORD  = local.email_credentials["smtp_password"],
```
alexwlchan, line 102: Not to address here, but I'd quite like us to get away from passing credentials in Lambda environment variables, and put them in Secrets Manager instead.

kenoir: I think this would require (a) adding permissions for this lambda to read credentials out of SM in the right place, and (b) a change to the lambda code to read from there and decode the creds.

I agree handing secrets around in this case is unsatisfactory.

PR #189, opened 2021-09-30
Email templates

gestchild:
Fixes wellcomecollection/wellcomecollection.org#6856
Fixes wellcomecollection/wellcomecollection.org#6951

Adds the code for email template customisation to deploy as part of the identity build.

Emails tested using Litmus in the following email clients:
litmus-large


universal_login/deploy.ts
```diff
@@ -160,13 +172,38 @@ async function updateLoginPageTemplate(
...
     console.log(`Updating environment: ${env}`);
 
     const token = await getApiToken(env);
     const prompts: Prompt[] = ['login', 'reset-password', 'email-verification'];
+    const emails: Email[] = [
+      'verify_email',
+      'welcome_email',
+      'blocked_account',
+      'change_password',
+    ];
```
jamesgorrie, line 206: You can create an array and derive the types from it.

e.g.
```ts
const emails = [
  'verify_email',
  'welcome_email',
  'blocked_account',
  'change_password',
] as const

type Email = typeof emails[number]
```

universal_login/emails/welcome_email.json
```diff
@@ -0,0 +1,8 @@
...
+  "from": "library@wellcomecollection.org",
+  "subject": "Welcome to Wellcome Collection",
+  "syntax": "liquid",
+  "enabled": true
+}
```
jamesgorrie, line 8: Where does this HTML get generated? Could we have the build script pull in an HTML file that's easier to edit, and construct the JSON from there to help with dev experience?

gestchild: I've put separate html files in for the email templates

jamesgorrie (approved):
Great!

PR #207, opened 2021-11-10
Add password validation to significant operations

jamieparkinson:
Some operations should require the user's current password to be validated in order for them to succeed. Rather than doing this in the client with 2 separate API calls, this PR makes it so that the relevant endpoints validate the password themselves.

### DIFF
--- a/packages/apps/api/src/handlers/user.ts
+++ b/packages/apps/api/src/handlers/user.ts
@@ -19,6 +19,7 @@ import { toSearchResults, toUser } from '../models/user';
 import { EmailClient } from '../utils/EmailClient';
 
 export function validatePassword(auth0Client: Auth0Client) {
+  const checkPassword = passwordCheckerForUser(auth0Client);
   return async function (request: Request, response: Response): Promise<void> {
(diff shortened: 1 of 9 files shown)


packages/apps/api/src/handlers/user.ts
```diff
@@ -287,28 +301,39 @@ export function changePassword(
...
     }
 
+    const auth0Get: APIResponse<Auth0Profile> = await auth0Client.getUserByUserId(
+      userId
+    );
+    if (auth0Get.status !== ResponseStatus.Success) {
+      throw clientResponseToHttpError(auth0Get);
+    }
+    const sss = extractSourceIp(request);
+    const pogs = await checkPassword(auth0Get.result.email, oldPassword, sss);
```
alexwlchan, line 324: What are these variable names?

alexwlchan (approved):

packages/apps/api/src/handlers/user.ts
jamieparkinson: oh no you've seen into my brain

PR #211, opened 2021-11-23
Remove admin stuff

jamieparkinson:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7344

We don't use the admin functionality, so it shouldn't be there: it hugely increases the vulnerable surface area of the API. 

This removes anything that looks like `isAdministrator` etc, and also the following routes:

- `GET /users` - user list/search
- `DELETE /users/:id` - user deletion
- `PUT /users/:id/reset-password` - manual reset password email sending (superseded by universal login)
- `PUT /users/:id/send-verification` - manual verification email sending
- `PUT, DELETE /users/:id/lock` - manual user locking and unlocking
- `DELETE /users/:id/deletion-request` - removal of a user's deletion request (creation is still possible)


jamesgorrie (approved):
🧹

PR #213, opened 2021-11-24
Legacy DB connection scripts

jamieparkinson:
This implements password and email change scripts, but leaves the verification script as a stub because we don't yet know how to do that in Sierra.


packages/apps/auth0-actions/src/helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+export const recordNumberForEmail = async (
+  email: string,
+  sierraClient: SierraClient
+): Promise<number | undefined> => {
+  const patronRecordResponse = await sierraClient.getPatronRecordByEmail(email);
+
+  if (patronRecordResponse.status === ResponseStatus.Success) {
+    return patronRecordResponse.result.recordNumber;
+  } else if (patronRecordResponse.status === ResponseStatus.NotFound) {
+    return undefined;
```
alexwlchan, line 22: I wonder if this should be an error also?

```suggestion
throw new Error(`Could not find user with email address [${email}]`)
```

So errors are returned consistently.

alexwlchan (approved):

packages/apps/auth0-actions/src/helpers.ts
jamieparkinson: You'd think so, but as per [the documentation of the callback values](https://auth0.com/docs/connections/database/custom-db/templates/change-email), we should be returning...
> a null first parameter with no corresponding second parameter (or one with a value of false) indicates that no password change was performed (possibly due to the user not being found)

alexwlchan: Could you put a comment to that effect/link to that documentation?

PR #215, opened 2021-11-25
Removing Sierra mutations from the API

jamieparkinson:
We do these in custom DB scripts now; some of the functionality they added has been moved into said scripts

### DIFF
--- a/packages/apps/auth0-actions/src/change_password.ts
+++ b/packages/apps/auth0-actions/src/change_password.ts
@@ -1,6 +1,6 @@
 import { callbackify } from 'util';
 import { HttpSierraClient } from '@weco/sierra-client';
-import { ResponseStatus } from '@weco/identity-common';
+import { ResponseStatus, truncate } from '@weco/identity-common';
 import { recordNumberForEmail } from './helpers';
 
 declare const configuration: {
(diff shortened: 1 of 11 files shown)


packages/apps/auth0-actions/src/change_password.ts
```diff
@@ -25,7 +25,8 @@ async function changePassword(
   if (recordNumber) {
     const patronRecordUpdate = await sierraClient.updatePassword(
       recordNumber,
-      newPassword
+      // Sierra ignores passwords after the 30th character
```
alexwlchan, line 28: Good find! Is this documented anywhere? Is it worth enforcing this in the Auth0 rules?

alexwlchan (approved):

alexwlchan:
To explain in a more concrete way: consider the following graph.

Screenshot 2021-11-25 at 16 33 26

We've seen four works matched into a five-node graph. Because one work is missing, the merger will output four distinct works.

I think you could reasonably argue that it should really merge A+B and D+E.

packages/apps/auth0-actions/src/change_password.ts
jamieparkinson: I'm not sure about this, because I just copied it from the existing implementation, which was introduced here https://github.com/wellcomecollection/identity/pull/40. Now we're keeping passwords in Sierra it seems more important, but I'm not sure what to do about it as I don't know how/what we'd communicate to users about it...

PR #217, opened 2021-11-29
Add resource server (aka custom API)

jamieparkinson:
This tells Auth0 that we will be creating a resource server (this will be in the custom authorizer) which will be able to validate tokens


alexwlchan (approved):
👍 Really helpful comments.

PR #222, opened 2021-12-02
Put the dev client authorizations into terraform

jamieparkinson:
So we don't accidentally blat these when applying

### DIFF
--- a/infra/scoped/auth0-resource-server.tf
+++ b/infra/scoped/auth0-resource-server.tf
@@ -44,3 +44,11 @@ resource "auth0_resource_server" "identity_api" {
   token_lifetime_for_web                          = 60 * 60 // 1 hour
   skip_consent_for_verifiable_first_party_clients = true
 }
+
+resource "auth0_client_grant" "dev_and_test" {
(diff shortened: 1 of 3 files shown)


infra/scoped/auth0-resource-server.tf
```diff
@@ -44,3 +44,11 @@ resource "auth0_resource_server" "identity_api" {
   token_lifetime_for_web                          = 60 * 60 // 1 hour
   skip_consent_for_verifiable_first_party_clients = true
 }
+
+resource "auth0_client_grant" "dev_and_test" {
+  for_each = toset(terraform.workspace != "prod" ? local.stage_test_client_ids : [])
```
alexwlchan, line 49: This seems like a weird way to phrase the ternary.
```suggestion
  for_each = toset(terraform.workspace == "staging" ? local.stage_test_client_ids : [])
```

alexwlchan (approved):

infra/scoped/auth0-resource-server.tf
jamieparkinson: Yeah, I copied this from elsewhere but I'll change both instances

gestchild (approved):

PR #226, opened 2021-12-07
Put the patron barcode in the app metadata

jamieparkinson:
This means that the API no longer talks to Sierra at all. Currently-logged-in users will need to log out and in again to see their barcode.

As part of this work, I removed an intermediate type which took data from Auth0 and renamed a few fields as there didn't seem to be any need for it. Most of the line count is in propagating that change through.


packages/apps/api-authorizer/src/authorization.ts
```diff
@@ -21,17 +22,6 @@ type ResourceRules = {
   };
 };
 
-export const userIdFromSubject = (
-  subjectClaim?: string
-): string | undefined => {
-  const prefix = 'auth0|p';
-  if (subjectClaim?.startsWith(prefix)) {
```
alexwlchan, line 28: I like how you've improved on the variable names in the new copy of this. 👍

packages/apps/api/src/app.ts
```diff
@@ -16,7 +16,6 @@ import { errorHandler } from './handlers/errorHandler';
 import { EmailClient } from './utils/EmailClient';
 
 export type Clients = {
-  sierra: SierraClient;
```
alexwlchan, line 19: Don't forget to remove your SierraClient import!

packages/apps/api/src/handlers/user.ts
```diff
@@ -18,15 +17,15 @@ export function validatePassword(auth0Client: Auth0Client) {
...
+    const auth0Get: APIResponse<Auth0User> = await auth0Client.getUserByUserId(
       userId
     );
     if (auth0Get.status !== ResponseStatus.Success) {
       throw clientResponseToHttpError(auth0Get);
     }
 
     await checkPassword(
-      auth0Get.result.email,
+      auth0Get.result.email!,
```
alexwlchan, line 28: Am I right in thinking the exclamation mark casts this to `undefined` if the result is `null`? If so: when would `auth0Get.result.email` be `null`?

packages/apps/api/src/server-lambda.ts
```diff
@@ -1,17 +1,10 @@
 import { APIGatewayProxyEvent, Context } from 'aws-lambda';
 import * as awsServerlessExpress from 'aws-serverless-express';
 import { Server } from 'http';
-import { HttpSierraClient, SierraClient } from '@weco/sierra-client';
 import { HttpAuth0Client, Auth0Client } from '@weco/auth0-client';
 import { createApplication } from './app';
 import HttpEmailClient, { EmailClient } from './utils/EmailClient';
 
-const sierraClient: SierraClient = new HttpSierraClient(
-  process.env.SIERRA_API_ROOT!,
```
alexwlchan, line 10: We need to remember to remove these from the Terraform definitions.

packages/apps/auth0-actions/src/helpers.ts
```diff
@@ -1,13 +1,16 @@
...
-export const patronRecordToUser = (patronRecord: PatronRecord): User => ({
+export const patronRecordToUser = (patronRecord: PatronRecord): Auth0User => ({
   user_id: 'p' + patronRecord.recordNumber,
   email: patronRecord.email,
   name: patronRecord.firstName + ' ' + patronRecord.lastName,
   given_name: patronRecord.firstName,
   family_name: patronRecord.lastName,
+  app_metadata: {
+    barcode: patronRecord.barcode,
+  },
```
alexwlchan, line 13: Yesssssssss

alexwlchan (approved):
🧴 Really nice work!

packages/apps/api/src/app.ts
jamieparkinson: oops

alexwlchan: delete even more code!!!!!

packages/apps/api/src/handlers/user.ts
jamieparkinson: `!` asserts that something that is typed as `T | undefined` is actually a `T` - for obvious reasons, the auth0 user type always has an `email` defined (I guess it's typed as `string | undefined` because some IDPs don't use emails?) so we make this assertion here.

Yesterday I convinced myself that it was preferable to assert at the point of use rather than at the module boundary, but I've changed my mind now and will do it in `Auth0Client`...

packages/apps/api/src/server-lambda.ts
jamieparkinson: Good point, I'll put that in this PR

PR #227, opened 2021-12-08
Use the correct name for database scripts

jamieparkinson:
In the process of figuring out the difference between Actions and Rules (answer: Actions are the replacement for Rules), I realised that this module is not named very well at all, because it doesn't contain any Actions.


packages/apps/auth0-database-scripts/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@weco/identity-auth0-actions",
+  "name": "@weco/auth0-database-scripts",
   "version": "0.1.0",
   "repository": "git@github.com:/wellcomecollection/identity.git",
-  "author": "Gary Tierney <gary.tierney@fastmail.com>",
```
alexwlchan, line 5: Why on earth was a personal email address here.

alexwlchan (approved):

PR #229, opened 2021-12-09
Add type definitions for post-login action trigger

jamieparkinson:
While the online editor has Typescript definitions built in, these don't seem to be publicly available - I've asked about this in the Auth0 forums: https://community.auth0.com/t/public-type-definitions-availability/74480.

In the mean time, each [trigger documentation page](https://auth0.com/docs/actions/triggers) contains a `Reference` section with details of the `Event` and `API` objects, so I've just replicated those docs here but in typescript form. Bit annoying, but at least they exist now.

### DIFF
--- a/packages/apps/auth0-actions/src/types/README.md
+++ b/packages/apps/auth0-actions/src/types/README.md
@@ -1,0 +1,7 @@
+# Actions environment type definitions
+
+While the online editor has Typescript definitions built in, these don't seem to be publicly available.
+
+We have asked about this in the Auth0 forums: https://community.auth0.com/t/public-type-definitions-availability/74480.
+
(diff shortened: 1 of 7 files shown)


packages/apps/auth0-actions/src/types/README.md
```diff
@@ -0,0 +1,7 @@
+# Actions environment type definitions
+
+While the online editor has Typescript definitions built in, these don't seem to be publicly available.
```
alexwlchan, line 3: nit

```suggestion
While the online editor has TypeScript definitions built in, these don't seem to be publicly available.
```

alexwlchan (approved):

PR #233, opened 2021-12-13
Tighten up session config

jamieparkinson:
Refresh token rotation is a cool feature: https://auth0.com/docs/security/tokens/refresh-tokens/refresh-token-rotation

I've tested this by temporarily setting access token lifetimes to 10 seconds (everything still works seamlessly, which is cool)

### DIFF
--- a/infra/scoped/locals.tf
+++ b/infra/scoped/locals.tf
@@ -103,6 +103,7 @@ locals {
   one_hour_s = 60 * 60
 
   // These should match the values used by the frontend app
+  // See: https://github.com/wellcomecollection/wellcomecollection.org/blob/f98e423c1ba75a703c9be07bba12c6060d55acab/identity/webapp/src/utility/auth0.ts#L80-L81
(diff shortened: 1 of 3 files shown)


infra/scoped/locals.tf
```diff
@@ -99,4 +99,10 @@ locals {
   # The hold limit was increased to 15 on 6 August 2021.
   # See https://wellcome.slack.com/archives/CUA669WHH/p1628089731008800
   per_user_hold_limit = 15
+
+  one_hour_s = 60 * 60
+
+  // These should match the values used by the frontend app
```
alexwlchan, line 105: Where would I find the values used by the frontend app?

alexwlchan (approved):

infra/scoped/locals.tf
jamieparkinson: I knew I should do this but couldn't bothered at the time 🙃

PR #237, opened 2021-12-15
Sort out dependencies

jamieparkinson:
This does these things:
- Removes all unnecessary dependencies from the workspace root
- Gives each workspace the correct dependencies and devDependencies
- Hoists all dependencies to the workspace root, except for those required in Lambda packaging
- Stops bundling dev dependencies in Lambdas, reducing the Lambda package size from 42mb to 1.2mb. This improves deployment speed.

### DIFF
--- a/packages/shared/sierra-client/tests/http-sierra-client.test.ts
+++ b/packages/shared/sierra-client/tests/http-sierra-client.test.ts
@@ -7,6 +7,7 @@ describe('HTTP sierra client', () => {
   let client: HttpSierraClient;
 
   beforeEach(() => {
+    // @ts-ignore
     moxios.install(axios as AxiosInstance);
     moxios.stubRequest(apiRoot + '/token', {
       status: 200,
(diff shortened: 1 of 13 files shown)


packages/shared/sierra-client/tests/http-sierra-client.test.ts
```diff
@@ -7,6 +7,7 @@ describe('HTTP sierra client', () => {
   let client: HttpSierraClient;
 
   beforeEach(() => {
+    // @ts-ignore
```
alexwlchan, line 10: What are we ignoring here? Can we be more specific in the annotations?

alexwlchan (approved):
🍠

packages/shared/sierra-client/tests/http-sierra-client.test.ts
jamieparkinson: I'm not sure - it should've been failing before but wasn't because the type dependencies weren't installed correctly, and I just copied it from the Auth0 Client test. I am planning to get rid of Moxios in my next PR so not too bothered about it...

PR #239, opened 2021-12-17
Replace moxios with MSW

jamieparkinson:
These tests leave a little to be desired imo but this is largely a like-for-like replacement rather than an improvement. This work is in service of upcoming changes to these clients which will mean they don't fetch a new token for every request.

### DIFF
--- a/packages/shared/auth0-client/tests/auth0-client.test.ts
+++ b/packages/shared/auth0-client/tests/auth0-client.test.ts
@@ -3,42 +3,29 @@ import {
   ResponseStatus,
   SuccessResponse,
 } from '@weco/identity-common';
-import axios, { AxiosInstance } from 'axios';
-import moxios from 'moxios';
 import { SierraUserIdPrefix, SierraConnection } from '../src/auth0';
 import { Auth0User, HttpAuth0Client } from '../src';
+import mockAuth0Server, { apiRoot, routeUrls } from './mock-auth0-server';
+import { compose, rest } from 'msw';
 
 describe('HTTP Auth0 client', () => {
-  let client: HttpAuth0Client;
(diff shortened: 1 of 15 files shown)


packages/shared/auth0-client/tests/auth0-client.test.ts
```diff
@@ -102,53 +70,50 @@ describe('HTTP Auth0 client', () => {
...
   });
 });
 
-const apiRoot: string = 'http://localhost';
-const apiAudience: string = 'http://my-api.localhost';
-const clientId: string = 'abcdefghijklmnopqrstuvwxyz';
-const clientSecret: string = 'ABCDEFGHIJKLMNOPQRSTUVYWXYZ';
-
-const accessToken: string = 'a1b23c4d5e6f7g8hj';
 const userId: number = 123456;
```
alexwlchan, line 117: Can we clean up these values now they're in `test-user.ts`?

packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -6,13 +6,7 @@ import {
...
   unhandledError,
 } from '@weco/identity-common';
-
-import {
-  extractRecordNumberFromLink,
-  PatronRecord,
-  toCreatePatron,
-  toPatronRecord,
-} from './patron';
+import { PatronRecord, toPatronRecord } from './patron';
```
alexwlchan, line 9: Why weren't these caught by linting previously?

alexwlchan (approved):

packages/shared/sierra-client/src/HttpSierraClient.ts
jamieparkinson: I don't know, I wondered the same!

PR #242, opened 2022-01-17
Map patron types into our own role definitions

jamieparkinson:
This maps the patron types from Sierra records into our own role-ish enum. "-ish", because "SelfRegistered" isn't really a role but as per discussion with @jtweed this is a pragmatic choice for integrating with Sierra and getting done what we need to be done.

The role in question gets stuck in the `app_metadata` and mapped into the SAML claims for use by OpenAthens

### DIFF
--- a/packages/shared/sierra-client/src/patron.ts
+++ b/packages/shared/sierra-client/src/patron.ts
@@ -1,10 +1,11 @@
-export function toPatronRecord(patronRecord: any): PatronRecord {
-  const patronName = getPatronName(patronRecord.varFields);
+export function toPatronRecord(patronResponse: any): PatronRecord {
+  const patronName = getPatronName(patronResponse.varFields);
   return {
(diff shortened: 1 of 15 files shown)


jtweed (approved):

packages/shared/sierra-client/src/patron.ts
```diff
@@ -96,17 +97,41 @@ function getPatronNameNonMarc(
   };
 }
 
-export interface PatronRecord {
+export type Role = 'Reader' | 'Staff' | 'SelfRegistered' | 'Excluded';
+
+// You can find an up-to-date list of patronType codes and descriptions
+// in the Sierra API at: /v5/patrons/metadata
```
alexwlchan, line 103: 👍

Useful comment.

alexwlchan (approved):

PR #245, opened 2022-01-27
(1) Verification notes creation

jamieparkinson:
This PR adds functions for creating and parsing verification status on/from the notes `varField`: it does not add the ability to do this into the Sierra client.

### DIFF
--- a/packages/shared/sierra-client/src/email-verification-notes.ts
+++ b/packages/shared/sierra-client/src/email-verification-notes.ts
@@ -30,6 +30,9 @@ const messageRegex = Object.values(messages).join('|');
 // - the prefix is followed by an email address (captured in group `email`)
 // - that is followed by one of the messages
 // - it ends with an ISO 8601 datetime string (captured in group `date`)
+//
+// For example, it would match:
+// `Auth0: example@example.com verified by the user clicking a verification email at 2022-02-22T00:00:00.000Z`
 const emailNoteRegex = new RegExp(
   `^${auth0NotePrefix}\\s+(?<email>${emailRegex})\\s+(?:${messageRegex})\\s+(?<date>${isoDateRegex})\$`
 );
--- a/packages/shared/sierra-client/src/email-verification-notes.ts
+++ b/packages/shared/sierra-client/src/email-verification-notes.ts
@@ -121,16 +121,3 @@ export const updateVerificationNote = (
 
   return addNotesVarfields(updatedNotes, varFields);
 };
-
-export const deleteNonCurrentVerificationNotes = (
-  varFields: VarField[]
-): VarField[] => {
-  const currentNotes = getVarFieldContent(varFields, varFieldTags.notes);
-  const email = getVarFieldContent(varFields, varFieldTags.email)[0] || '';
-  const updatedNotes = currentNotes.filter((note) => {
-    const verified = parseVerificationNote(note);
-    return !verified || verified.email === email;
-  });
-
-  return addNotesVarfields(updatedNotes, varFields);
-};
--- a/packages/shared/sierra-client/src/email-verification-notes.ts
+++ b/packages/shared/sierra-client/src/email-verification-notes.ts
@@ -12,11 +12,12 @@ const isoDateRegex =
   '\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d+(?:[+-][0-2]\\d:[0-5]\\d|Z)';
 
 // The distinction here is that we assume that users who exist already in Sierra
-// are verified, because they must've used their registered email address to set a password.
+// are explicitlyVerified, because they must've used their registered email address to set a password.
 // We distinguish between them in the message to maximise readability and to ease debugging.
 const messages = {
-  verified: 'verified at',
-  assumedVerified: 'assumed to be verified at login',
+  explicitlyVerified: 'verified by the user clicking a verification email at',
+  implicitlyVerified:
+    'implicitly verified on initial Auth0 login with an existing Sierra account at',
 };
 
 const messageRegex = Object.values(messages).join('|');
@@ -33,6 +34,10 @@ const emailNoteRegex = new RegExp(
 type VerificationNote = {
   email: string;
   date: Date;
+};
+
+type NoteOptions = {
+  type: 'Implicit' | 'Explicit';
 };
 
 const parseVerificationNote = (note: string): VerificationNote | undefined => {
@@ -52,16 +57,18 @@ const parseVerificationNote = (note: string): VerificationNote | undefined => {
 
 const createVerificationNote = (
   { email, date = new Date() }: VerificationNote,
-  verificationWasAssumed: boolean = false
+  { type }: NoteOptions
 ): string =>
   [
     auth0NotePrefix,
     email,
-    verificationWasAssumed ? messages.assumedVerified : messages.verified,
+    type === 'Implicit'
+      ? messages.implicitlyVerified
+      : messages.explicitlyVerified,
     date.toISOString(),
   ].join(' ');
 
-const filterVerificationNotes = (notes: string[]): string[] =>
+const removeVerificationNotes = (notes: string[]): string[] =>
   notes.filter((note) => {
     const parsedNote = parseVerificationNote(note);
     return !parsedNote;
@@ -69,11 +76,11 @@ const filterVerificationNotes = (notes: string[]): string[] =>
 
 const upsertVerificationNote = (
   note: VerificationNote,
-  verificationWasAssumed: boolean = false,
+  opts: NoteOptions,
   notes: string[]
 ): string[] => [
-  ...filterVerificationNotes(notes),
-  createVerificationNote(note, verificationWasAssumed),
+  ...removeVerificationNotes(notes),
+  createVerificationNote(note, opts),
 ];
 
 const addNotesVarfields = (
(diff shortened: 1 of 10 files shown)


packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,128 @@
+import { getVarFieldContent, VarField, varFieldTags } from './patron';
+
+const auth0NotePrefix = 'Auth0:';
+
+// See https://stackoverflow.com/a/9204568
+// Matches string@string.string
+const emailRegex = '\\S+@\\S+\\.\\S+';
```
alexwlchan, line 7: Aargh regex!

I've had a look through all the patron email addresses currently in Sierra. There are a handful that don't match this regex, of which the only one I think we need to care about is https://trustnet.wellcome.org/user/odonnelo.

The others are either clearly nonsense (e.g. `NONE`) or values with trailing/leading whitespace, that we should remove or we won't be able to look them up.

Maybe worth a comment that (1) we know this is an oversimplification and (2) we're okay with it, because we only have a small-ish pool of values.

packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,131 @@
...
+
+// See https://stackoverflow.com/a/3143231
+// Matches ISO 8601 dates like YYYY-MM-DDTHH:MM:SS.SSSSZ
+const isoDateRegex =
+  '\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d+(?:[+-][0-2]\\d:[0-5]\\d|Z)';
+
+// The distinction here is that we assume that users who exist already in Sierra
+// are verified, because they must've used their registered email address to set a password.
+// We distinguish between them in the message to maximise readability and to ease debugging.
```
alexwlchan, line 19: 👍 Good idea.

packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,131 @@
...
+  verificationWasAssumed: boolean = false
+): string =>
+  [
+    auth0NotePrefix,
+    email,
+    verificationWasAssumed ? messages.assumedVerified : messages.verified,
+    date.toISOString(),
+  ].join(' ');
+
+const filterVerificationNotes = (notes: string[]): string[] =>
```
alexwlchan, line 67: Is this filtering *for* verification notes, or removing them?

Maybe `removeVerificationNotes`?

packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  return addNotesVarfields(updatedNotes, varFields);
+};
+
+export const deleteNonCurrentVerificationNotes = (
+  varFields: VarField[]
+): VarField[] => {
+  const currentNotes = getVarFieldContent(varFields, varFieldTags.notes);
+  const email = getVarFieldContent(varFields, varFieldTags.email)[0] || '';
+  console.log(email);
```
alexwlchan, line 124: Is this leftover debugging code?

packages/shared/sierra-client/tests/email-verification-notes.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        {
+          fieldTag: varFieldTags.barcode,
+          content: '1234567',
+        },
+      ]);
+
+      expect(result).toHaveLength(3);
+      expect(
+        result.find(({ fieldTag }) => fieldTag === varFieldTags.notes)?.content
+      ).toBe('Auth0: example@example.com verified at 2022-02-22T00:00:00.000Z');
```
alexwlchan, line 40: I'd throw in an assertion that both the original notes are preserved; this test would pass if we put in three copies of our note!

packages/shared/sierra-client/tests/email-verification-notes.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      jest.useFakeTimers();
+      jest.setSystemTime(new Date('2022-02-22'));
+      const result = addVerificationNote(
+        [
+          {
+            fieldTag: varFieldTags.email,
+            content: 'example@example.com',
+          },
+        ],
+        true
```
alexwlchan, line 53: Can we assign this to a named variable? I hate seeing naked booleans.

packages/shared/sierra-client/tests/email-verification-notes.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  beforeAll(() => {
+    jest.useFakeTimers();
+    jest.setSystemTime(new Date('2022-02-22'));
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
+
+  describe('addVerificationNote', () => {
```
alexwlchan, line 24: What if we add a verification note and there's already one there? What if the existing note has the same email address? A different email address?

alexwlchan (approved):
Looks good, just a suggestion about some more test cases we might want to cover.

packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,131 @@
...
+// Matches ISO 8601 dates like YYYY-MM-DDTHH:MM:SS.SSSSZ
+const isoDateRegex =
+  '\\d{4}-[01]\\d-[0-3]\\dT[0-2]\\d:[0-5]\\d:[0-5]\\d\\.\\d+(?:[+-][0-2]\\d:[0-5]\\d|Z)';
+
+// The distinction here is that we assume that users who exist already in Sierra
+// are verified, because they must've used their registered email address to set a password.
+// We distinguish between them in the message to maximise readability and to ease debugging.
+const messages = {
+  verified: 'verified at',
+  assumedVerified: 'assumed to be verified at login',
```
alexwlchan, line 22: Sorry to bikeshed, but looking through subsequent PRs I'm falling out of love with this "assumed" naming. The further away you get, the less obvious it is what this means. Let's make it super obvious so these notes make sense even when this code is long and gone.

Maybe we could call it explicit/implicit verification? Then you get:

```typescript
const messages = {
  explicitVerification: 'explicitly verified by the user at',
  implicitVerification: 'implicitly verified on initial Auth0 login with an existing Sierra account at',
}
```

jamieparkinson: Yeah, I think I discounted this naming at some point but actually it looks fine now - will change that

packages/shared/sierra-client/tests/email-verification-notes.test.ts
jamieparkinson: I didn't test for these because it just does what it says - it adds one and doesn't care about the others

alexwlchan: \*nod* – I think that's being covered in tests in a subsequent PR.

alexwlchan: Okay, no, I'm confused again.

I guess maybe this is a design question I want to clarify: when a user changes their email address, do we scrub the previous verification notes, or do we keep them? It looks like maybe we only scrub them on their next login, is that correct?

jamieparkinson: Your understanding is correct, and answering this question is the sole reason that this PR took so long to come up with. 

If you scrub the previous verification note when the user changes their email address, then your user becomes indistinguishable from a user who's appropriate for "implicit verification" - so they can verify themselves just by logging in again, which is wrong. 

However, now that I think about it again having done some more work on this, I think there might be another way to do it. 
- When a user changes their email address, do nothing to their notes (we can see that they're not verified because their email does not match their verification note)
- When a user verifies their email address, add the new verification note and delete the old one.
- When a user logs in, add the (implicit) verification note if they have _no_ verification notes.

I do remember trying this and discounting it, but perhaps something else changed after I did this - not sure. Can you see any problems with it?

alexwlchan: \*nod* Yup, I think I was suggesting closer to the second flow – when I said "changes their email address", I really meant "verifies their email address after changing it". The second flow sounds fine to me, but now I'm trying to think if there's a hole.

The only wrinkle I can think of (but it applies both way) is a well-meaning staff member deleting a verification note for an old email address, and getting us into the implicit flow. But that seems quite unlikely.

packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -0,0 +1,123 @@
...
+export const verifiedEmail = (varFields: VarField[]): string | undefined =>
+  getVarFieldContent(varFields, varFieldTags.notes)
+    .map(parseVerificationNote)
+    .find((verified): verified is VerificationNote =>
+      Boolean(verified && verified.date.getTime() <= Date.now())
+    )?.email;
+
+export const updateVerificationNote = (
+  varFields: VarField[],
+  opts: NoteOptions = { type: 'Explicit' }
```
alexwlchan, line 109: Do we need a default here? Is there any case we wouldn't pass a value for `opts`?

packages/shared/sierra-client/src/patron.ts
```diff
@@ -120,13 +131,31 @@ const patronTypeToRole = (patronType: number): Role => {
   }
 };
 
+export const varFieldTags = {
+  barcode: 'b',
+  email: 'z',
+  notes: 'x',
+  name: 'n',
+  message: 'm',
+} as const;
```
alexwlchan, line 140: This is really nice.

alexwlchan (approved):

packages/shared/sierra-client/src/email-verification-notes.ts
jamieparkinson: I quite like the default hiding the distinction for the "usual" use case - the implicit verification is an oddity that I would like to keep to only the places that need to know about it

PR #246, opened 2022-01-27
(2) Add verification notes to Sierra client

jamieparkinson:
This adds the ability to mark the patron's current email verified

### DIFF
--- a/packages/shared/sierra-client/src/HttpSierraClient.ts
+++ b/packages/shared/sierra-client/src/HttpSierraClient.ts
@@ -7,8 +7,12 @@ import {
   successResponse,
   unhandledError,
 } from '@weco/identity-common';
-import { PatronRecord, toPatronRecord } from './patron';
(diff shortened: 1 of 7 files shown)


packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -131,44 +135,160 @@ export default class HttpSierraClient implements SierraClient {
...
+              error
+            );
+          case 404:
+            return errorResponse(
+              'Patron record with record number [' +
+                recordNumber +
+                '] not found',
+              ResponseStatus.NotFound,
+              error
+            );
```
alexwlchan, line 185: Not for this patch, but it occurs to me we might want to consolidate some of this logic so we don't have to repeat it on all responses.

alexwlchan (approved):

alexwlchan (approved):

PR #249, opened 2022-01-31
We only need to patch the notes varfields

jamieparkinson:
The way that the Sierra patron update API works is a bit odd - you pass it an array, but it's non-destructive and it only actually takes heed of notes, messages, and patron image fields. Because of this, we only actually need to patch the notes varFields: this is very convenient, because it sidesteps an issue where an existing patron image field would cause any updates to fail due to the lack of support for this feature.

### DIFF
--- a/packages/shared/sierra-client/src/HttpSierraClient.ts
+++ b/packages/shared/sierra-client/src/HttpSierraClient.ts
@@ -168,7 +168,7 @@ export default class HttpSierraClient implements SierraClient {
           // varField with empty content. As such, passing an empty array
           // is a no-op (there's nothing to add). This is observed
           // behaviour, which is quite vaguely documented:
-          // 
+          //
           // https://techdocs.iii.com/sierraapi/Content/zObjects/requestObjectDescriptions.htm#patronPUT
           [];
 
--- a/packages/shared/sierra-client/src/HttpSierraClient.ts
+++ b/packages/shared/sierra-client/src/HttpSierraClient.ts
@@ -159,8 +159,17 @@ export default class HttpSierraClient implements SierraClient {
 
       const notesVarFields = verified
         ? updateVerificationNote(currentRecordResponse.data.varFields)
-        : // The way that varFields patches work in the Sierra API means
-          // that PUTting an empty array will be a no-op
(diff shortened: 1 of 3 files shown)


packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -152,15 +157,17 @@ export default class HttpSierraClient implements SierraClient {
         }
       );
 
-      const updatedVarFields = verified
+      const notesVarFields = verified
         ? updateVerificationNote(currentRecordResponse.data.varFields)
-        : currentRecordResponse.data.varFields;
+        : // The way that varFields patches work in the Sierra API means
+          // that PUTting an empty array will be a no-op
```
alexwlchan, line 163: Is this documented behaviour, or is it something we've observed?

It might be worth recording the distinction, so that if it changes we spot it and don't start deleting notes inadvertently.

alexwlchan (commented):
Is there a document that explains how/why we use notes for email verification?

I think it'd be useful, and a good place to document this slightly unusual behaviour – because if it ever stops behaving this way, we'll start blatting notes on records.

packages/shared/sierra-client/src/HttpSierraClient.ts
jamieparkinson: Yeah, good point

packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -152,15 +157,26 @@ export default class HttpSierraClient implements SierraClient {
...
+      const notesVarFields = verified
         ? updateVerificationNote(currentRecordResponse.data.varFields)
-        : currentRecordResponse.data.varFields;
+        : // The Sierra API's behaviour re varFields is non-intuitive:
+          // we pass an array of varFields that we want to _add_. These
+          // have to be x, m, or y varFields (notes, messages, images)
+          // and anything else gets ignored.
+          //
+          // If we want to remove one, we do this by passing the
+          // varField with empty content. As such, passing an empty array
```
alexwlchan, line 168: This is even more bizarre than I realised. Good comment, thank you!

alexwlchan (approved):
🇸🇱

PR #250, opened 2022-02-08
Add the infra and skeleton app for Patron deletion tracking

jamieparkinson:
This is fairly predictable boilerplate for a CI-deployed TypeScript Lambda in this repo - some of the tf structure isn't ideal but this fits the existing pattern.

The diff is mostly because I blatted and recreated the `yarn.lock`; this also created some minor TS changes.

### DIFF
--- a/packages/apps/api-authorizer/src/authorization.ts
+++ b/packages/apps/api-authorizer/src/authorization.ts
@@ -26,11 +26,19 @@ export const hasScopes = (...requiredScopes: string[]): AccessControlRule => (
   jwt,
   parameters
 ) => {
+  if (typeof jwt.payload === 'string') {
+    return false;
+  }
+
   const tokenScopes: string[] = jwt.payload.scope?.split(' ') || [];
   return tokenScopes.some((tokenScope) => requiredScopes.includes(tokenScope));
 };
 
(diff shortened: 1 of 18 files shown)


packages/apps/api-authorizer/src/authorization.ts
```diff
@@ -26,11 +26,19 @@ export const hasScopes = (...requiredScopes: string[]): AccessControlRule => (
   jwt,
   parameters
 ) => {
+  if (typeof jwt.payload === 'string') {
```
paul-butcher, line 29: Why might jwt.payload be a string at this point, and why would we not process it?

Also same question in AccessControlRule, below.

jamieparkinson: This change came from my blatting of `yarn.lock` which led to a semver-compatible upgrade of `@types/jsonwebtoken` where opaque/non-JSON payloads were (correctly) [represented](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58158) - I'll add a comment to this effect as it's not especially clear

paul-butcher (approved):
LGTM

alexwlchan (approved):
Maybe worth a README with a few sentences explaining what this app does?

PR #252, opened 2022-02-10
Add a method to the Sierra client for fetching deleted patrons

jamieparkinson:
Does what it says

### DIFF
--- a/packages/shared/sierra-client/src/HttpSierraClient.ts
+++ b/packages/shared/sierra-client/src/HttpSierraClient.ts
@@ -110,6 +110,45 @@ export default class HttpSierraClient implements SierraClient {
     });
   }
 
+  async getDeletedRecordNumbers({
+    start,
+    end,
+  }: {
+    start?: Date;
+    end?: Date;
+  } = {}): Promise<APIResponse<number[]>> {
+    try {
+      const instance = await this.getInstance();
+      const startDateString = start?.toISOString().slice(0, 10);
+      const endDateString = (end ?? new Date()).toISOString().slice(0, 10);
+
+      const deletedDate = startDateString
+        ? `[${startDateString}, ${endDateString}]`
+        : undefined;
+
+      const response = await instance.get<{ entries: Array<{ id: number }> }>(
+        '/patrons',
+        {
+          params: {
+            deleted: true,
+            deletedDate,
+          },
+          validateStatus: (status) => status === 200,
+        }
+      );
+      return successResponse(response.data.entries.map((entry) => entry.id));
+    } catch (error) {
+      if (error.response) {
+        switch (error.response.status) {
+          // Sierra returns a 404 rather than an empty list
+          case 404:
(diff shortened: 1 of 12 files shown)


packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -110,6 +110,45 @@ export default class HttpSierraClient implements SierraClient {
...
+      const response = await instance.get<{ entries: Array<{ id: number }> }>(
+        '/patrons',
+        {
+          params: {
+            deleted: true,
+            deletedDate,
+          },
+          validateStatus: (status) => status === 200,
+        }
+      );
```
alexwlchan, line 138: What if there are more deletions than fit in a single page? Will this correctly paginate over them, or does it just get the first page?

jamieparkinson: I wonder how many times I have written, and how many times more I will write, a wrapper for a paginated API

packages/shared/sierra-client/src/HttpSierraClient.ts
```diff
@@ -110,6 +111,47 @@ export default class HttpSierraClient implements SierraClient {
...
+    end?: Date;
+  } = {}): Promise<APIResponse<number[]>> {
+    try {
+      const instance = await this.getInstance();
+      const startDateString = start?.toISOString().slice(0, 10);
+      const endDateString = (end ?? new Date()).toISOString().slice(0, 10);
+
+      const deletedDate = startDateString
+        ? `[${startDateString}, ${endDateString}]`
+        : undefined;
```
alexwlchan, line 128: If an end date isn't specified by the user, do we need to supply one here, or could we just drop in an empty string?

e.g. `GET /patrons?deletedDate=[2001-01-01,]`

packages/shared/sierra-client/src/pagination.ts
```diff
@@ -0,0 +1,45 @@
...
+      params: {
+        ...(config.params ?? {}),
+        limit: pageLimit,
+        id: typeof lastId !== 'undefined' ? `[${lastId + 1},]` : undefined,
+      },
+    });
+
+    const entries = response.status === 404 ? [] : response.data.entries;
+    if (entries.length >= pageLimit) {
+      return getResults(entries[entries.length - 1].id, [
```
alexwlchan, line 36: minor readability suggestion:

```suggestion
      const lastId = entries[entries.length - 1].id
      return getResults(lastId, [
```

alexwlchan (approved):

packages/shared/sierra-client/src/HttpSierraClient.ts
jamieparkinson: See above and/or the tests - it defaults to the current date. I know the Sierra API is (sometimes) OK with half-bounded intervals but I'd prefer to be explicit.

PR #254, opened 2022-02-21
Implement the deletion tracking lambda

jamieparkinson:
Most of the lines here are tests - it was a real pain to test!

**What this PR does**:
- Implements a Lambda that looks for yesterday's deleted patrons in Sierra and deletes them in Auth0
- Respects the Auth0 API's rate limiting
- Gives good warnings if something goes wrong

**What this PR doesn't do** (but future PRs will):
- Wire up the Lambda to report errors to Slack
- Look at windows other than the previous day - this will become configurable via the event.

### DIFF
--- a/packages/apps/patron-deletion-tracker/src/app.ts
+++ b/packages/apps/patron-deletion-tracker/src/app.ts
@@ -1,7 +1,62 @@
 import { SierraClient } from '@weco/sierra-client';
 import { Auth0Client } from '@weco/auth0-client';
+import { ResponseStatus } from '@weco/identity-common';
+import { ScheduledHandler } from 'aws-lambda';
+import { startOfYesterday, endOfYesterday } from 'date-fns';
+import { rateLimit } from './rate-limits';
+
+const auth0RateLimited = rateLimit({
+  // See https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/management-api-endpoint-rate-limits#self-service-subscription-limits
+  rateLimits: {
+    minute: 200,
+    second: 20,
+  },
+  maxConcurrency: 5,
+});
+
+// The time before the Lambda will exit at which we should log an error that we
+// won't be able to complete the deletions
+const lambdaExitThreshold = 2000;
 
 export const createApp = (
   auth0Client: Auth0Client,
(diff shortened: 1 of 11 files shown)


packages/apps/patron-deletion-tracker/src/app.ts
```diff
@@ -1,5 +1,60 @@
...
+import { Auth0Client } from '@weco/auth0-client';
+import { ResponseStatus } from '@weco/identity-common';
 import { ScheduledHandler } from 'aws-lambda';
+import { startOfYesterday, endOfYesterday } from 'date-fns';
+import { rateLimit } from './rate-limits';
 
-export const lambdaHandler: ScheduledHandler = async (event) => {
-  console.log('I am a Patron deletion tracker');
+const auth0RateLimited = rateLimit({
+  // See https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/management-api-endpoint-rate-limits#self-service-subscription-limits
```
alexwlchan, line 9: It'd be helpful to explain which endpoint we're using – although I can see it's "Write users" now because it's the only endpoint with these numbers, if those change it'll be harder to tell.

Also, is max concurrency set by us or by Auth0?

packages/apps/patron-deletion-tracker/src/app.ts
```diff
@@ -1,5 +1,60 @@
...
+  rateLimits: {
+    minute: 200,
+    second: 20,
+  },
+  maxConcurrency: 5,
+});
+
+// The time before the Lambda will exit at which we should log an error that we
+// won't be able to complete the deletions
```
alexwlchan, line 18: Note: a Lambda timeout becomes an error that goes to the Lambda's DLQ, and our alerting will catch it.

alexwlchan (approved):
It might be worth getting somebody who knows more TS to look over this; it looks fine to me but the tests look pretty complicated and I may be missing stuff.

packages/apps/patron-deletion-tracker/src/app.ts
jamieparkinson: Will expand this comment to include that the intention is around logging the actual failed records

PR #255, opened 2022-02-25
Be more permissive when parsing verification notes

jamieparkinson:
If we can extract meaning from them, that's good enough. The fully readable string and the date are conveniences on top of a field which fundamentally just needs to store a verified email address.

### DIFF
--- a/packages/shared/sierra-client/src/email-verification-notes.ts
+++ b/packages/shared/sierra-client/src/email-verification-notes.ts
@@ -28,18 +28,23 @@ const messageRegex = Object.values(messages).join('|');
 // This matches a note where:
 // - it begins with the prefix (Auth0:)
 // - the prefix is followed by an email address (captured in group `email`)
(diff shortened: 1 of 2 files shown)


packages/shared/sierra-client/src/email-verification-notes.ts
```diff
@@ -54,11 +59,11 @@ const parseVerificationNote = (note: string): VerificationNote | undefined => {
 
   const { email, date: dateString } = match.groups;
   const date = new Date(dateString);
-  if (!email || isNaN(date.getDate())) {
+  if (!email) {
     return undefined;
   }
 
-  return { email, date };
+  return { email, date: isNaN(date.getDate()) ? undefined : date };
```
alexwlchan, line 66: minor: move this into the `date` variable above?

alexwlchan (approved):
…well that lasted 😂 

Good idea, good comments.

PR #256, opened 2022-02-25
Make email templates a bit more configurable

jamieparkinson:
Currently the staging emails had to have the same `resultUrl` as the prod emails, which was a bit of a pain - this makes them more configurable, and refactors the deploy script across a few files

### DIFF
--- a/universal_login/deploy.ts
+++ b/universal_login/deploy.ts
@@ -1,13 +1,9 @@
 #!/usr/bin/env ts-node-script
-
-import AWS from 'aws-sdk';
 import fs from 'fs';
-import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
+import { AxiosInstance } from 'axios';
+import { RateLimiter } from 'limiter';
+import { getAuthenticatedInstance } from './scripts/auth0-management';
 
-const secretsManager = new AWS.SecretsManager();
-const ssm = new AWS.SSM();
-
-type Env = 'stage' | 'prod';
 type Template = 'universal-login';
 const prompts = ['login', 'reset-password', 'email-verification'] as const;
 const emails = [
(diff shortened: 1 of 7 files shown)


universal_login/deploy.ts
```diff
@@ -64,164 +79,45 @@ function loadEmail(emailName: Email): Record<string, unknown> {
...
+): Promise<void> {
+  const emailData = loadEmail(email);
+  const emailEndpoint = `/api/v2/email-templates/${email}`;
 
-    for (const prompt of prompts) {
-      await delay();
-      console.log(`Updating ${prompt} prompt`);
-      await updateTextPrompt(env, prompt, token);
+  for (const [key, value] of Object.entries(emailData)) {
+    if (typeof value === 'string') {
```
alexwlchan, line 114: Is it worth logging a warning if we get a non-string value, or could that never happen?

universal_login/deploy.ts
```diff
@@ -64,164 +79,45 @@ function loadEmail(emailName: Email): Record<string, unknown> {
...
+  return instance.put(emailEndpoint, emailData);
+}
 
-    console.log('Updates completed successfully');
-  } catch (e) {
-    console.error(e);
-    process.exit(1);
-  }
-})();
+const env = process.argv[2] as Env;
```
alexwlchan, line 122: What sort of error do we throw if `process.argv[2]` is missing? Or it's not an `Env` value?

I'd add a `try … catch` here (or the TS equivalent) that prints a usage string if it's used incorrectly.

alexwlchan (approved):
🎊

universal_login/deploy.ts
jamieparkinson: `Record<string, string>` should prevent that, TS doesn't seem to be able to reconcile that with `Object.entries`

PR #257, opened 2022-02-28
Wire the deletion tracking lambda up to the Lambda alerting service

jamieparkinson:
Errors -> Slack


alexwlchan (approved):
We should really have a proper module for Lambdas.

(Maybe once we get round to logging to Logstash.)

PR #261, opened 2022-03-09
Make patron deletion windows wholly configurable

jamieparkinson:
There is going to be a local script for triggering deletion checks over the whole window that we're interested in, but I'm waiting until I can publish https://www.npmjs.com/package/@weco/ts-aws before I implement that

### DIFF
--- a/packages/apps/patron-deletion-tracker/src/windows.ts
+++ b/packages/apps/patron-deletion-tracker/src/windows.ts
@@ -1,0 +1,42 @@
+import {
+  addDays,
+  endOfDay,
+  endOfYesterday,
+  startOfDay,
+  startOfToday,
+  subDays,
+} from 'date-fns';
+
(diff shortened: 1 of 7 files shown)


packages/apps/patron-deletion-tracker/src/windows.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  addDays,
+  endOfDay,
+  endOfYesterday,
+  startOfDay,
+  startOfToday,
+  subDays,
+} from 'date-fns';
```
melanierogan, line 8: i like date-fns :)

melanierogan (approved):

packages/apps/patron-deletion-tracker/src/windows.ts
jamieparkinson: Yeah it's nice isn't it!

PR #266, opened 2022-03-09
Add a couple more Staff patron types

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/7787

### DIFF
--- a/packages/shared/sierra-client/tests/patron.test.ts
+++ b/packages/shared/sierra-client/tests/patron.test.ts
@@ -126,7 +126,7 @@ describe('toPatronRecord', () => {
   });
 
   it('throws an error for unexpected patron types', () => {
-    const record = createRecordWithPatronType(2); // Research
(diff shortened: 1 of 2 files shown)


packages/shared/sierra-client/tests/patron.test.ts
```diff
@@ -126,7 +126,7 @@ describe('toPatronRecord', () => {
   });
 
   it('throws an error for unexpected patron types', () => {
-    const record = createRecordWithPatronType(15); // Inter Lib Loans
+    const record = createRecordWithPatronType(2); // Research
```
alexwlchan, line 129: It doesn't matter too much, but I'd be tempted to put a number we don't actually use for a patron type here, to prevent this test starting to fail again, e.g.

```suggestion
    const record = createRecordWithPatronType(1000); // Higher than all known patron types
```

alexwlchan (approved):

PR #267, opened 2022-03-14
Add ability to invoke deletion tracker from a script

jamieparkinson:
This does what it says on the tin, and also fixes a few bugs in the deletion tracker. Don't be afraid of the diff - it's almost all `yarn.lock`!

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -13,7 +13,9 @@
     "target": "es2017",
     "baseUrl": "./packages/shared",
     "paths": {
-      "@weco/*": ["./*/src"]
+      "@weco/auth0-client": ["./auth0-client/src"],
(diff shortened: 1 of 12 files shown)


tsconfig.base.json
```diff
@@ -13,7 +13,9 @@
     "target": "es2017",
     "baseUrl": "./packages/shared",
     "paths": {
-      "@weco/*": ["./*/src"]
+      "@weco/auth0-client": ["./auth0-client/src"],
+      "@weco/sierra-client": ["./sierra-client/src"],
+      "@weco/identity-common": ["./identity-common/src"]
```
jamieparkinson, line 18: This change is necessary because of our `@weco`-namespaced npm package!

melanierogan (approved):

PR #269, opened 2022-03-17
Lambda error metric alarms have a naming convention

jamieparkinson:
See: https://github.com/wellcomecollection/platform-infrastructure/blob/a2d2ba64212a6d2f78570d73b6345b287c0517a3/monitoring/slack_alerts/lambda_errors_to_slack_alerts/src/lambda_errors_to_slack_alerts.py#L135-L139


melanierogan (approved):
learned something myself here, thanks!

jamieparkinson:
Yeah, this should be documented or (better) enforced, but right now I just want to Get Things Working

PR #271, opened 2022-03-17
Drop the email template deployment rate limit further

jamieparkinson:
Saw errors related to this, hopefully will fix it


universal_login/deploy.ts
```diff
@@ -59,7 +59,7 @@ const doDeploy = async (env: Env) => {
 // 2 req/second - this ensures we never hit that
 // https://auth0.com/docs/troubleshoot/customer-support/operational-policies/rate-limit-policy/management-api-endpoint-rate-limits
 const rateLimiter = new RateLimiter({
-  tokensPerInterval: 2,
+  tokensPerInterval: 1,
```
alexwlchan, line 62: Let's update the comment to explain that we've deliberately lowered it from the Auth0 threshold, so a well-meaning person doesn't come back to "fix" this later.

alexwlchan (approved):

PR #275, opened 2022-03-21
Fix/login duplicate email

melanierogan:
## Problem
As a user
When I enter an email address that is duplicated
I want a helpful error message so I'm clear why something has gone wrong

Ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/7594

## What has changed
I have tried to allow for the Error status 409/Duplicate patron in `get_user`, then add a case for it in `login`. We handle 409 Errors with `ResponseStatus.MalformedRequest`, I have adjusted our custom `duplicateEmailCredentialMessage` to throw that error at point of login if Sierra gives us a Malformed Request at login. 

## Further context
In auth0 logs, this error appears as:
`Unhandled API response [Request failed with status code 409] (cause: [{"code":133,"specificCode":0,"httpStatus":409,"name":"Internal server error","description":"Duplicate patrons found for the specified varFieldTag`

### DIFF
--- a/packages/apps/auth0-database-scripts/src/get_user.ts
+++ b/packages/apps/auth0-database-scripts/src/get_user.ts
@@ -25,8 +25,8 @@ async function getUser(email: string): Promise<Auth0User | undefined> {
   } else if (patronRecordResponse.status === ResponseStatus.NotFound) {
     return undefined;
   } else if (patronRecordResponse.status === ResponseStatus.MalformedRequest) {
-    //Sierra returns status 409 for duplicate patrons, our error handling calls this a MalformedRequest
-    //here we return undefined to fail silently to progress to login where we will catch the duplicate error
+    // Sierra returns status 409 for duplicate patrons, our error handling calls this a MalformedRequest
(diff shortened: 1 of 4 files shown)


jamieparkinson:
Nice detective work on this - have you thought about how it'll affect other flows apart from login?

As per [the docs](https://auth0.com/docs/authenticate/database-connections/custom-db/templates/get-user), the "Get User" script runs in the following situations: create user, change email, change password, and password reset. We also know it runs at login.

I think there is still a problem with the Change Email script - what happens if a user tries to change their email to one that is present in multiple records? I think(?) the call to `updatePatronEmail` will succeed, leaving the user unable to log in. I guess a fix could be to check for the existence of the user manually in the Change Email script, but then we're duplicating functionality - and, maybe more importantly, Sierra API calls - that Auth0 does for us.

melanierogan:
I'm going to close this PR as I have changed the approach to this particular ticket: ultimately I am going to focus on changing the default 'something went wrong, please try again later' wording. I want to refer to this closed PR in the ticket.

PR #281, opened 2022-03-28
Limit concurrency for identity deployment

jamieparkinson:
Stops errors when 2 PRs are merged around the same time


melanierogan (approved):
Had a read of the [buildkite docs](https://buildkite.com/docs/pipelines/controlling-concurrency) here to familiarise myself with concurrency limits, _Setting a concurrency limit of 1 on a step in your pipeline will ensure that no two jobs created from that step will run at the same time, even if there are agents available._

PR #285, opened 2022-03-31
Minimise errors relating to changing/reseting password 

melanierogan:
My [previous approach](https://github.com/wellcomecollection/identity/pull/282) to adding a description and note telling users about trivial/repeated pattern passwords wasn't going to work. This time I have gone about improving error in HttpAuth0Client and surfacing a further improved error message at point of error for repeated patterns in changed passwords.

This relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/7816

### DIFF
--- a/packages/shared/auth0-client/src/HttpAuth0Client.ts
+++ b/packages/shared/auth0-client/src/HttpAuth0Client.ts
@@ -210,6 +210,18 @@ export default class HttpAuth0Client implements Auth0Client {
                     ResponseStatus.PasswordTooWeak,
                     error
                   );
+                } else if (
+                  error.response.data?.message?.code === 136 &&
+                  error.responses.data?.message?.description.startsWith(
+                    'PIN is not valid : PIN is trivial'
+                  )
+                ) {
+                  return errorResponse(
+                    'Password contains repeated characters',
+                    ResponseStatus.PasswordTooWeak,
+                    error
+                  );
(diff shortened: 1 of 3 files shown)


packages/shared/auth0-client/src/HttpAuth0Client.ts
```diff
@@ -210,6 +210,17 @@ export default class HttpAuth0Client implements Auth0Client {
                     ResponseStatus.PasswordTooWeak,
                     error
                   );
+                } else if (
+                  error.response.data?.message?.code === 136 &&
+                  error.responses.data?.message?.description.startsWith(
+                    'PIN is not valid : PIN is trivial'
+                  )
```
jamieparkinson, line 217: Is there somewhere in the Sierra docs we can link to in a comment for (a) the code and (b) the Sierra definition of triviality?

jamieparkinson (approved):

packages/shared/auth0-client/src/HttpAuth0Client.ts
melanierogan: Would you believe me if I said there doesn't seem to be a Sierra definition of triviality? I can link to the [general error codes,](https://techdocs.iii.com/sierraapi/Content/zAppendix/errorHandling.htm?Highlight=error) which includes 136 but there doesn't seem to be a set section for the specific codes within that or of 'PIN is trivial'.

PR #286, opened 2022-03-31
Remove model validation from API Gateway

jamieparkinson:
🗑️


melanierogan (commented):
A really basic question from me, apologies: why did we have model validation in the API Gateway?

jamieparkinson:
The original API as was handed over to us by Digirati made use of much more of the APIGW feature set in lieu of application code than it does now, and also placed heavier emphasis on validation rather as opposed to error handling - we've moved away from those approaches now.

davidpmccormick (approved):

PR #291, opened 2022-04-14
Update custom claims to support roles

donaldgray:
Scope `weco:patron_role` returns `https://wellcomecollection.org/patron_role` containing roles, if present.


jamieparkinson (approved):
✨

PR #293, opened 2022-05-04
Create updateUserAfterRegistration endpoint

davidpmccormick:

Opted to create a new endpoint and handler instead of wrapping `updateUser` in conditionals based on whether we've got a password or not.

I think the `user_id` we're going to pass from the NextJS app will be prefixed with `auth0|` and I imagine we'll need to strip this off in order for the [`getTargetUserId` function](https://github.com/wellcomecollection/identity/blob/main/packages/apps/api/src/handlers/user.ts#L217-L238) not to blow up. I've seen elsewhere we're checking [if it starts with `auth0|p`](https://github.com/wellcomecollection/identity/blob/main/packages/shared/auth0-client/src/auth0.ts#L4) – I'm not sure where that `p` comes from (and it doesn't appear in the ids that I've been testing with)

__Edit:__ [I've removed the `auth0|` (and maybe `p`) before it gets this far](https://github.com/wellcomecollection/wellcomecollection.org/pull/7944/files#diff-5785a90503c26d63f092dff4b182b0f28a4a7375107aaddbfc9007997b2cd6fcR30-R31). I guess the `p` is for 'patron', although I wasn't sure if it would be added to newly created Sierra-database records using Auth0 as a way to signup, so left it optional for now. If we know the prefix either definitely will or will not contain a `p`, we can remove the `?` or the `p` from the regex, respectively.

### DIFF
--- a/infra/scoped/auth0-resource-server.tf
+++ b/infra/scoped/auth0-resource-server.tf
@@ -39,6 +39,11 @@ resource "auth0_resource_server" "identity_api" {
     description = "Update password"
   }
 
+  scopes {
+    value       = "update:user"
+    description = "Update user info"
+  }
+
   // Applications are allowed refresh tokens
   allow_offline_access                            = true
   token_lifetime                                  = 60 * 60 // 1 hour
--- a/packages/apps/api-authorizer/src/handler.ts
+++ b/packages/apps/api-authorizer/src/handler.ts
@@ -40,6 +40,9 @@ const validateRequest = resourceAuthorizationValidator({
     POST: allOf(isSelf, hasScopes('create:requests')),
     GET: allOf(isSelf, hasScopes('read:requests')),
   },
+  '/users/{userId}/registration': {
+    PUT: allOf(isSelf, hasScopes('update:user')),
+  },
 });
 
 export const createLambdaHandler = (validateToken: TokenValidator) => async (
--- a/packages/apps/api/src/app.ts
+++ b/packages/apps/api/src/app.ts
@@ -40,6 +40,9 @@ export function createApplication(clients: Clients): Application {
   return app;
 }
 
+// The handler for this endpoint duplicates a lot of the work in the handler for
+// `/users/:user_id` This is deliberate as 1. they may diverge and 2. we can do
+//  a check that the user is at the correct stage of the signup process here
 function registerUsersUserIdRegistrationResource(
   clients: Clients,
   app: Application
--- a/packages/apps/api/src/app.ts
+++ b/packages/apps/api/src/app.ts
@@ -9,6 +9,7 @@ import {
   getUser,
   requestDelete,
   updateUser,
+  updateUserAfterRegistration,
   validatePassword,
 } from './handlers/user';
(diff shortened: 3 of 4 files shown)


melanierogan (approved):
Looking good David. I'm just trying to wrap up the required methods for patron creation on HttpSierraClient right now, i'll have a think about `| p` I'm assuming on patron creation

jamieparkinson:
Can you add a comment about why this is a new endpoint? Reading the PR, I was initially against it as it's not very RESTful at all and it's definitely duplication, but I changed my mind because I thought (a) it seems reasonable that these endpoints will diverge (eg, we might want to send transactional emails here at some point in future) and (b) we can do a check that the user is at the correct stage of the signup process here.

jamieparkinson:
Further thoughts:

- The ID will always include a `p` - [this](https://github.com/wellcomecollection/identity/blob/36582841bb02bc17cad440851bd1b78c7017ae80/packages/shared/auth0-client/src/auth0.ts#L6-L12) might be some useful context if you look at where it's used?
- You need to let the authorizer know about this endpoint [here](https://github.com/wellcomecollection/identity/blob/38fcf0c7c0c8c590f528a31b0d6f7be5456af1c1/packages/apps/api-authorizer/src/handler.ts#L25-L43)
- This will also need some terraform additions in API gateway for it to work (can do in a separate PR). It's very boilerplatey:
  1. For a new route you need an `aws_api_gateway_resource`, and then for each method (eg `PUT`) that might be used you need a `aws_api_gateway_method`, and for each response that a given method can have you need a `aws_api_gateway_method_response`. Example for `OPTIONS /users/:id` [here](https://github.com/wellcomecollection/identity/blob/3ad3dba225a45bd893ae8fb2c5cfa80041a5f6ae/infra/scoped/api-gateway.tf#L34-L68).
  2. To connect the resource to the API you need an `aws_api_gateway_integration` for each method. Example for `GET /users/:id` [here](https://github.com/wellcomecollection/identity/blob/9344d5c3b4d8c05b5691f764cfb57d54daafc121/infra/scoped/api-integration.tf#L13-L27).
  3. You also need to add stuff for documentation, but in the course of writing this I think I've decided we should delete the documentation...

davidpmccormick:
> The ID will always include a p - [this](https://github.com/wellcomecollection/identity/blob/36582841bb02bc17cad440851bd1b78c7017ae80/packages/shared/auth0-client/src/auth0.ts#L6-L12) might be some useful context if you look at where it's used?

Does it make sense to send the whole prefixed id (e.g. `auth0|p12345678`) through to this app and remove the prefix here, or strip it before it gets here, just sending `12345678`?

jamieparkinson:
This endpoint needs to match the others in the API - the `auth0|p` prefix is private, and the routes should only include the plain `1234567`. The authorizer won't work if it's any other way.

davidpmccormick:
I _think_ I've addressed the comments with the exception of the API gateway stuff (which I think I'd prefer to do and understand in a separate PR).

jamieparkinson (approved):
Lovely!

PR #295, opened 2022-05-09
Adding createPatron method for sierra http client 

melanierogan:
Some tidying up to do

### Who is this for?

So we have an available sierra-http-client method to create a patron via `POST`

###  Ticket 

Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/7896 but specifically https://github.com/wellcomecollection/wellcomecollection.org/issues/7956
So I can link up with David's work on weco-identity and in the other identity PR for the new endpoint to create a patron on sierra from the formData

### DIFF
--- a/packages/shared/sierra-client/src/HttpSierraClient.ts
+++ b/packages/shared/sierra-client/src/HttpSierraClient.ts
@@ -80,6 +80,10 @@ export default class HttpSierraClient implements SierraClient {
       try {
         const response = await instance.post('/patrons/', {
           // Create the patron in marc format
+          // Rationale: After a short discussion with Natalie on what we ideally want from a newly created patron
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):

jamieparkinson:
This makes me realise that somewhere - inclined to think not in this client - we should be doing some sanity checking on the name/email fields to prevent misuse.

melanierogan:
@jamieparkinson I have changed approach on the `POST` params a little here to make it clear we are creating the patron in MARC format. I decided on this approach after speaking to Natalie on what the ideal situation would be for a new patron record: If a patron can be registered with a fully correct MARC record at the point of submitting the form, it leaves less work for library staff.

jamieparkinson:
Can you put that rationale in a comment?

PR #296, opened 2022-05-11
Post registration api gateway

davidpmccormick:

Add API gateway required things (I think).

### DIFF
--- a/infra/scoped/api-gateway.tf
+++ b/infra/scoped/api-gateway.tf
@@ -295,6 +295,176 @@ resource "aws_api_gateway_method_response" "users_userid_put_500" {
   }
 }
 
+# /users/:user_id/registration
+resource "aws_api_gateway_resource" "users_userid_registration" {
+  rest_api_id = aws_api_gateway_rest_api.identity.id
+  parent_id   = aws_api_gateway_resource.users_userid.id
+  path_part   = "registration"
+}
+
+# [OPTIONS]
+
+resource "aws_api_gateway_method" "users_userid_registration_options" {
+  rest_api_id   = aws_api_gateway_rest_api.identity.id
+  resource_id   = aws_api_gateway_resource.users_userid_registration.id
+  http_method   = "OPTIONS"
+  authorization = "NONE"
+}
+
+# 204 No Content
+
+resource "aws_api_gateway_method_response" "users_userid_registration_options_204" {
+  rest_api_id = aws_api_gateway_rest_api.identity.id
+  resource_id = aws_api_gateway_resource.users_userid_registration.id
+  http_method = aws_api_gateway_method.users_userid_registration_options.http_method
+  status_code = "204"
+
+  response_models = {
+    "application/json" = "Empty"
+  }
+
+  response_parameters = {
+    "method.response.header.Access-Control-Allow-Headers" = true,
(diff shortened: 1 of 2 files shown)


melanierogan (approved):
I think this looks right, but you might want a more seasoned terraformer to cast their eye too

infra/scoped/api-gateway.tf
```diff
@@ -295,6 +295,176 @@ resource "aws_api_gateway_method_response" "users_userid_put_500" {
...
+  }
+
+  response_parameters = {
+    "method.response.header.Access-Control-Allow-Origin" = true
+  }
+}
+
+# 400 Bad Request
+
+resource "aws_api_gateway_method_response" "users_userid_registration_put_400" {
```
alexwlchan, line 368: We have modules for configuring these error responses; see https://github.com/wellcomecollection/terraform-aws-api-gateway-responses

In particular, they'll set up the JSON error responses we use in the rest of our APIs, e.g.

```json
{
  "errorType": "http",
  "httpStatus": "400",
  "label": "Bad Request",
  "description": "You didn't combobulate the widget wrangler",
  "type":"Error"
}
```

For an internal-only API this is likely overkill and this is fine as-is, but I mention it for information.

infra/scoped/api-integration.tf
```diff
@@ -70,6 +70,35 @@ resource "aws_api_gateway_integration" "users_userid_password_put" {
...
 }
 
+# [OPTIONS] /users/:user_id/registration
+
+resource "aws_api_gateway_integration" "users_userid_registration_options" {
+  rest_api_id = aws_api_gateway_rest_api.identity.id
+  resource_id = aws_api_gateway_resource.users_userid_registration.id
+  http_method = aws_api_gateway_method.users_userid_registration_options.http_method
+
+  integration_http_method = "POST"
```
alexwlchan, line 80: Why is this POST when the comment says OPTIONS?

alexwlchan (approved):

infra/scoped/api-gateway.tf
jamieparkinson: fwiw - I probably wouldn't use that here, because it's not used for the rest of this APIGW setup (this is a biiiig file, which isn't great but it _is_ easy to copy and paste - and delete). Better would be to refactor the whole stack to be more modular in a separate piece of work.

alexwlchan: Yup, agreed – consistency within this stack is better than using this module.

jamieparkinson (approved):
hoops: jumped

infra/scoped/api-integration.tf
davidpmccormick: Good question – this was broadly copy/paste from the [`users/:user_id/password` block above](https://github.com/wellcomecollection/identity/pull/296/files/6312d9d7d63ab39b12738f7405e263f47679da19#diff-f13e44ae1ed19905dd828a72ee53102577e546d82a2dcb1ec87600dd2bc67d7aR45-R71), where it is also `POST` under the same circumstances. Do both of these need updating?

davidpmccormick: Great. Thanks for clarifying

PR #297, opened 2022-05-16
Add updateNames method to the Auth0Client

davidpmccormick:

Save the user's names to Auth0 after registration completion


jamieparkinson (commented):
Why a new method rather than changing `updateUser` to be parametrised with something like `{ email, firstName, lastName }`?

PR #300, opened 2022-05-27
Fix/update redirect action

melanierogan:
I need to update the auth0 redirect action to reflect the new variables names, and also allow for the fact that we don't need to decipher stage or prod values for redirect values in the action, we do this in tf.

### DIFF
--- a/packages/apps/auth0-actions/tests/redirect_to_full_registration.test.ts
+++ b/packages/apps/auth0-actions/tests/redirect_to_full_registration.test.ts
@@ -30,8 +30,8 @@ describe('redirect_to_full_registration', () => {
   it('redirects the user if terms_and_conditions_accepted is true and firstname, surname is present', () => {
     jest.clearAllMocks();
     const auth0SecretsObject = {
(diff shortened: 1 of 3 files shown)


packages/apps/auth0-actions/tests/redirect_to_full_registration.test.ts
```diff
@@ -119,8 +118,8 @@ describe('redirect_to_full_registration', () => {
         request: { hostname: 'stage.wellcomecollection.org' } as EventRequest,
         secrets: auth0SecretsObject as EventSecrets,
       } as Event<Auth0User>);
-    const envUrl = event.secrets.AUTH0_ACTION_URL_STAGE;
-    const successUrl = `https://${envUrl}/success`;
+
+    const successUrl = `https://${event.secrets.IDENTITY_APP_BASEURL}/success`;
```
alexwlchan, line 122: Why do we need to prepend the https here but not on line 60?

melanierogan: good catch! We don't, we just need `/success' - fix incoming 👍

jamieparkinson (approved):

PR #313, opened 2022-06-08
Signup form

davidpmccormick:
Part of #7812

Link to signup form from login form. Hide this link until we know you've got the `selfRegistration` toggle turned on
 
image

Signup form also links back to sign in form

image


universal_login/prompt/login.json
```diff
@@ -4,36 +4,7 @@
     "description": " ",
     "buttonText": "Sign in",
     "custom-script-error-code": "There is an issue with this library account. To resolve this, please contact Library Enquiries (library@wellcomecollection.org).",
-    "title": "title",
```
alexwlchan, line 7: Why have you deleted all these values? What were they being used for?

alexwlchan (approved):
📽️

universal_login/prompt/login.json
davidpmccormick: The two specific values that we wanted weren't showing up, so I put all available options on the page to see what rendered. The issue ended up being further upstream – having `disable signups` set at the connection level prevented the 'Sign up' text/link displaying.

So this is just to delete all the test values added one or two prs previous.

PR #315, opened 2022-06-09
Use query string to determine showSignup

davidpmccormick:
The toggle cookies are set on `wellcomecollection.org` so won't be available for the login screens at `account.wellcomecollection.org`.

Reading a query param (`showRegistration=true`) out of the URL instead to accomplish the same thing.


jamieparkinson (approved):
Do I understand right that this is just for the UI and the actual ability to sign up will remain switched off on prod regardless?

PR #316, opened 2022-06-13
Check the host contains 'stage' to determine showing signup link

davidpmccormick:
It seems like we should be able to [encode/decode arbitrary data in the `state` parameter](https://github.com/auth0/nextjs-auth0/blob/a7a588d651eee571e7aa8ee881fbabbcd61153d6/src/auth0-session/hooks/get-login-state.ts#L35) on the Auth0 hosted form screens, but I haven't been able to make this work. For now I've resorted to checking the host is stage.


melanierogan (approved):
I think this is good for now, wary of sinking too much time in if we can't encode/decode state 👍

PR #319, opened 2022-06-14
update the get user script

melanierogan:
We are testing patron creation through auth0 'create' script. We notice that we are getting a particular error which suggests the json is invalid. 

```Malformed or invalid Patron creation request (cause: [{\"code\":115,\"specificCode\":0,\"httpStatus\":400,\"name\":\"Invalid JSON\",\"description\":\"Invalid JSON : field(s) unknown : params.\"}])```

I have a feeling this will return undefined values for certain fields we are trying to populate in the POST create patron. If we just return, I'm hoping we get the values we need. 

FYI The flow of auth0 scripts
1. Get user script
2. Create user script
3. Login script

### DIFF
--- a/packages/apps/auth0-database-scripts/src/get_user.ts
+++ b/packages/apps/auth0-database-scripts/src/get_user.ts
@@ -23,7 +23,9 @@ async function getUser(email: string): Promise<Auth0User | undefined> {
     const patronRecord = patronRecordResponse.result;
     return patronRecordToUser(patronRecord);
   } else if (patronRecordResponse.status === ResponseStatus.NotFound) {
-    return undefined;
+    // If a user is not found in sierra, we should be able to create the user in
+    // the next script that runs, 'create'
+    return;
   } else {
     throw new Error(patronRecordResponse.message);
   }

### CODE REVIEW

File: packages/apps/auth0-database-scripts/src/get_user.ts [hunk 1]
```diff
@@ -23,7 +23,9 @@ async function getUser(email: string): Promise<Auth0User | undefined> {
     const patronRecord = patronRecordResponse.result;
     return patronRecordToUser(patronRecord);
   } else if (patronRecordResponse.status === ResponseStatus.NotFound) {
-    return undefined;
+    // If a user is not found in sierra, we should be able to create the user in
+    // the next script that runs, 'create'
+    return;
```
davidpmccormick, line 28: Will this not implicitly still be `undefined`? I'm willing to give it a shot though.

melanierogan: I think you could be right, I think i've misread the get user function and thought it'll pass on the auth0 profile to create

PR #321, opened 2022-06-15
adding pin to createPatron

melanierogan:
A user needs their password to be set on sierra, we can now do this at point of patron creation in create script. 

**Updated Note**: Based on a nifty suggestion by David I've added a new Auth0 type that contains password. 

**Note:** ~I've temporarily put a ts-ignore in for `user.password` as I can't see how to type this within the current existing type for Auth0User shown below, if anyone has any guidance on how best to type in this context it would be greatly appreciated. [Auth0 docs](https://auth0.com/docs/authenticate/database-connections/custom-db/templates/create#user-object-example) suggest that password is available from Auth0User.~

```
 export type Auth0User = WithRequiredFields<
  User<AppMetadata, UserMetadata>,
  'user_id' | 'email' // These fields are always present
>;
```

related to work on https://github.com/wellcomecollection/wellcomecollection.org/issues/8071, https://github.com/wellcomecollection/wellcomecollection.org/issues/8072 and https://github.com/wellcomecollection/wellcomecollection.org/issues/7896

### DIFF
--- a/packages/shared/sierra-client/tests/http-sierra-client.test.ts
+++ b/packages/shared/sierra-client/tests/http-sierra-client.test.ts
@@ -55,6 +55,7 @@ describe('HTTP sierra client', () => {
       lastName: 'Ravioli',
       firstName: 'Ravi',
       email: 'raviravioli@pastatimestest.com',
+      password: '12345abcdefg',
     };
     mockSierraServer.use(
       rest.post(routeUrls.patron, (req, res, ctx) => res(ctx.json(newPatron)))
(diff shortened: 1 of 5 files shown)


packages/shared/sierra-client/tests/http-sierra-client.test.ts
```diff
@@ -64,7 +65,8 @@ describe('HTTP sierra client', () => {
       const response = await client.createPatron(
         'Ravioli',
         'Ravi',
-        'raviravioli@ppastatimestest.com'
+        'raviravioli@ppastatimestest.com',
```
davidpmccormick, line 68: looks like ravi has typo-ed their email?

melanierogan: That's a bad pasta move eh?

PR #323, opened 2022-06-15
Use specific errors for each case

davidpmccormick:


packages/apps/auth0-database-scripts/src/login.ts
```diff
@@ -10,8 +10,16 @@ declare const configuration: {
   CLIENT_SECRET: string;
 };
 
-const invalidCredentialsMessage =
-  "We don't recognise the email and/or password you entered. Please check your entry and try again.";
+// const invalidCredentialsMessage =
+//   "We don't recognise the email and/or password you entered. Please check your entry and try again.";
```
jamieparkinson, line 14: why comment?

davidpmccormick: I think I figured we'd be putting this back in if/when we figure out where the problem is. I guess there's no harm in removing it altogether for now though.

melanierogan (approved):
✨

PR #326, opened 2022-06-16
implement a generic updatePatron method 

melanierogan:
This can take options like barcode and varFieldTags to change patron details via PUT. 

Based on https://github.com/wellcomecollection/wellcomecollection.org/issues/8073 and suggestion from @jamieparkinson  here https://github.com/wellcomecollection/wellcomecollection.org/issues/8073#issuecomment-1156277878
and overall related to https://github.com/wellcomecollection/wellcomecollection.org/issues/7896

### DIFF
--- a/packages/apps/auth0-database-scripts/src/change_password.ts
+++ b/packages/apps/auth0-database-scripts/src/change_password.ts
@@ -25,10 +25,10 @@ async function changePassword(
   const recordNumber = await recordNumberForEmail(email, sierraClient);
 
   if (recordNumber) {
-    const patronRecordUpdate = await sierraClient.updatePassword(
+    const patronRecordUpdate = await sierraClient.updatePatron(
       recordNumber,
       // Sierra ignores passwords after the 30th character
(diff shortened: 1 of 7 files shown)


jamieparkinson (commented):
How hard would it be to roll together the other mutator methods?
- `updatePatronEmail`
- `updatePassword`

I think `markPatronEmailVerified` is sufficiently different in behaviour that it can/should stay separate

melanierogan:
> How hard would it be to roll together the other mutator methods?
> 
> * `updatePatronEmail`
> * `updatePassword`
> 
> I think `markPatronEmailVerified` is sufficiently different in behaviour that it can/should stay separate

I think password pretty quick and patron email i'll just have to have another look at but can't immediately see why not. Please hold - more code deletion to come 🎵

melanierogan:
I've got `updatePassword` rolled in, `updateEmail` is a little more tricky as it GETs record number, then updates an email and adds a note.

PR #333, opened 2022-06-21
More logging

melanierogan:
Logging out whole `APIGatewayRequestAuthorizerEvent` temporarily, also, check for undefined token ~and throw error if token is undefined~ and log it


packages/apps/api-authorizer/src/handler.ts
```diff
@@ -52,6 +52,10 @@ export const createLambdaHandler = (validateToken: TokenValidator) => async (
   const authHeader = event.headers?.['Authorization'];
   const token = authHeader?.match(/^Bearer (.*)$/)?.[1];
   console.log(token, '<<<<< TOKEN VALUE');
+  console.log(event, '<<<<< WHOLE API GATEWAY REQUEST AUTHORIZER EVENT');
+  if (token == undefined) {
+    throw new Error('Token is undefined');
```
alexwlchan, line 57: We want a log rather than an error, so we still send the correct HTTP 401 response to the client.

PR #334, opened 2022-06-21
Point to the correct url for the registration form

davidpmccormick:
☝️


alexwlchan (approved):

alexwlchan:
@davidpmccormick This is a change I applied manually, but it'll get rolled back on the next deploy. Can you sort out the failing tests?

PR #335, opened 2022-06-21
Comments to make debugging this stuff easier

alexwlchan:
Partly these are for my own notes; partly these are because Auth0/Buildkite was being distinctly weird earlier and we couldn't tell if it was actually updating the database action scripts or now.

Prepending the date and commit should make it easier to spot if/when Buildkite is updating the scripts, and when something is stuck.

### DIFF
--- a/.buildkite/scripts/auth0-deployment.sh
+++ b/.buildkite/scripts/auth0-deployment.sh
@@ -29,10 +29,17 @@ function __do_deployment() {
   done
 
   # Deploy actions
+  #
+  # We prepend the date when the script was written to help debug issues
+  # with the deployment.
   for action in ${ACTIONS}
   do
     action_name=$(jq -r ".names.${action}" ${AUTH0_ACTIONS_FILE})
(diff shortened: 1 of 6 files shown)


.buildkite/scripts/auth0-deployment.sh
```diff
@@ -19,14 +27,21 @@ function __do_deployment() {
   # Deploy database scripts
   for script in ${DATABASE_SCRIPTS}
   do
-    cp -v "${SCRIPTS_BUILD_DIR}/${script}.js" "${AUTH0_EXPORT_DIR}/database-connections/${AUTH0_CONNECTION_NAME}/"
+    dest_script="${AUTH0_EXPORT_DIR}/database-connections/${AUTH0_CONNECTION_NAME}/${script}.js"
+
+    cp -v "${SCRIPTS_BUILD_DIR}/${script}.js" "$dest_script"
+    echo -e "$DEBUG_HEADER\n\n$(cat "$dest_script")" > "$dest_script"
```
jamieparkinson, line 33: let's use the webpack plugin that we already have for this; as well as being much more readable it means that commit hashes are associated with the build itself rather than the point at which the build is deployed (the latter could become misleading)

jamieparkinson (approved):

PR #340, opened 2022-06-23
update patron first and lastname

melanierogan:
Once we get the first and lastname through to Identity api registration endpoint we need to be able to update their record on Sierra with the incoming first and lastname.

### DIFF
--- a/packages/apps/api/src/handlers/user.ts
+++ b/packages/apps/api/src/handlers/user.ts
@@ -77,17 +77,8 @@ export function updateUserAfterRegistration(
       throw clientResponseToHttpError(auth0Update);
     }
 
-    // Get the patron recordNumber
-    const getPatronRecordResponse = await sierraClient.getPatronRecordByEmail(
-      auth0Profile.email
-    );
-    if (getPatronRecordResponse.status !== ResponseStatus.Success) {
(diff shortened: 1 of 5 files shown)


packages/apps/api/src/handlers/user.ts
```diff
@@ -76,7 +77,34 @@ export function updateUserAfterRegistration(
...
 
-    // TODO: Update patron firstName and lastName in sierra
+    // Get the patron recordNumber
+    const getPatronRecordResponse = await sierraClient.getPatronRecordByEmail(
+      auth0Profile.email
+    );
+    if (getPatronRecordResponse.status !== ResponseStatus.Success) {
+      throw clientResponseToHttpError(getPatronRecordResponse);
+    }
+    const { recordNumber } = getPatronRecordResponse.result;
```
jamieparkinson, line 87: Is this call necessary? I believe the `userId` _is_ the `recordNumber`

jamieparkinson (approved):
Great - fyi if the way the test fixtures (`MockThingClient`s) are annoying then feel free to change how they work, I wrote them in an (exceedingly rare) burst of productivity and I'm not sure that they're very good...

PR #342, opened 2022-06-24
Only allow machine users to call the registration endpoint

alexwlchan:


melanierogan (approved):
Really appreciate the clear notes on the approach

packages/apps/api-authorizer/src/authorization.ts
```diff
@@ -12,7 +12,9 @@ type HttpVerb =
   | 'CONNECT';
 
 type AccessControlRule = (
-  jwt: Jwt,
+  jwt: Jwt & {
+    payload: (JwtPayload & { gty?: string }) | string
```
jamieparkinson, line 16: I don't think we need both checks and `gty` seems more hacky that the `@client` suffix (which is, at least, inside a meaningful claim) - can we just use the latter?

alexwlchan: Sure, changed.

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2020 Wellcome Collection
+Copyright (c) 2020-2022 Wellcome Collection
```
jamieparkinson, line 3: I could never have guessed that there is [no agreed correct way](https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year) to do something straightforward with a FOSS license

jamieparkinson (approved):

LICENSE
alexwlchan: Oh, haha, I just did this to trigger the pre-commit hook, but apparently it only passes modified files to prettier. 🤷‍♀️

PR #346, opened 2022-06-24
Replace pre-commit with husky and lint-staged

jamieparkinson:
Was a bit of a pain to sort out the CI side of this but I got there in the end - it turns out there were a load of things in `main` which were not correctly prettiered (I'm not sure how this happened), so this PR also includes those autoformatting changes.


jamieparkinson:
The files were added after the prettier hook was set up - my best guess is that the change is with an increased prettier version somewhere!

alexwlchan (approved):

PR #348, opened 2022-06-27
Check a user is mid-registration when updating their name

### LINKED ISSUE #347: Only allow updating the first/last name in the registration endpoint for a partially-registered user
The registration endpoint in the identity API is very powerful; it can update the first and last name of *any* user in Sierra. If somebody found a way to abuse this endpoint, it would be bad!

This endpoint should only be used when after a user is signing up for the first time, which means their account will be in a known state. In particular, they'll have the placeholder name we give to all new users:

Screenshot 2022-06-24 at 15 03 26

To give us an extra level of protection, the identity API endpoint should look for this placeholder name, and refuse to update a patron with a different name.

alexwlchan:
Closes #347

Inside the `PUT /users/{userId}/registration` endpoint, we now retrieve the existing patron record from Sierra before updating the first/last name.

This gives us some additional security on this endpoint: in particular, we'll reject updates for a patron if they don't have our Auth0 placeholder name `Auth0_Registration_undefined Auth0_Registration_tempLastName`.

This is the intended logic:

```mermaid
flowchart TD
    GET{get patron<br/>from Sierra}

    GET -- does not exist --> E404[HTTP 404<br/>Not Found]
    GET -- other error --> E500[HTTP 500<br/>Internal Server Error]

    GET --> MATCH{does stored<br/>name match<br/>the request?}

    MATCH -- yes --> NOOP[no-op]
    MATCH -- no --> PLACE{is stored name<br/>the placeholder?}
    
    PLACE -- yes --> UPDATE[update name in Sierra<br/>HTTP 204 No Content]
    PLACE -- no --> E409[HTTP 409 Conflict]
    

    classDef decisionNode fill:#e8e8e8,stroke:#8f8f8f
    class GET,PLACE,MATCH decisionNode

    classDef errorNode fill:#f8c4c9,stroke:#e01b2f,stroke-width:2px
    class E404,E500,E409 errorNode

    classDef successNode fill:#d1ffe9,stroke:#007868,stroke-width:2px
    class NOOP,UPDATE successNode
```

I've also changed how the UpdateUser process works slightly:

* We can't update the name in Auth0 and I don't think we need to, so we don't
* We need to send a list of varfields to Sierra, not a single one

### DIFF
--- a/packages/apps/api/test/routes/users_user_id_registration.test.ts
+++ b/packages/apps/api/test/routes/users_user_id_registration.test.ts
@@ -74,7 +74,7 @@ describe('/users/{userId}/registration', () => {
           .send({ firstName, lastName })
       );
 
-      expect(response.statusCode).toBe(400);
+      expect(response.statusCode).toBe(409);
 
       const sierraUser = await clients.sierra.getPatronRecordByRecordNumber(
         testUser.userId
--- a/packages/apps/api/test/routes/users_user_id_registration.test.ts
+++ b/packages/apps/api/test/routes/users_user_id_registration.test.ts
@@ -62,7 +62,10 @@ describe('/users/{userId}/registration', () => {
     it('409s if a user is already registered', async () => {
       const firstName = 'Silas';
       const lastName = 'Burroughs';
-      const testUser = randomExistingUser({ firstName: 'Henry', lastName: 'Wellcome' });
+      const testUser = randomExistingUser({
+        firstName: 'Henry',
+        lastName: 'Wellcome',
+      });
       const { api, clients } = mockedApi([testUser]);
 
       const response = await withCallerId('@machine')(
--- a/packages/apps/api/test/routes/users_user_id_registration.test.ts
+++ b/packages/apps/api/test/routes/users_user_id_registration.test.ts
@@ -59,7 +59,7 @@ describe('/users/{userId}/registration', () => {
       }
     });
 
-    it('400s if a user is already registered', async () => {
+    it('409s if a user is already registered', async () => {
       const firstName = 'Silas';
       const lastName = 'Burroughs';
       const testUser = randomExistingUser({ firstName: 'Henry', lastName: 'Wellcome' });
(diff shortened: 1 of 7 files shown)


packages/apps/api/test/routes/users_user_id_registration.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      const sierraUser = await clients.sierra.getPatronRecordByRecordNumber(
+        testUser.userId
+      );
+
+      if (sierraUser.status === ResponseStatus.Success) {
+        expect(sierraUser.result.firstName).toBe(firstName);
+        expect(sierraUser.result.lastName).toBe(lastName);
+      } else {
+        throw new Error(`Unexpected failure from Sierra: ${sierraUser}`);
```
jamieparkinson, line 32: Nit: I would rather not throw "documentary" errors like this in tests; rather I would think about the assertions/expectations that we want and then use the [expect DSL](https://jestjs.io/docs/expect) to express them. In some projects we've used https://github.com/jest-community/jest-extended to expand our vocabulary there.

jamieparkinson (approved):
This is really great Alex - I am a huge believer in using the guard-clause procedural style for security-critical code. Also pleased to see the test fixtures have held up so well to a new use case!

packages/apps/api/test/routes/users_user_id_registration.test.ts
alexwlchan: Yeah, I don't love this – it's a bit of a hack around the type system. Here `sierraUser` is `SuccessResponse | ErrorResponse`, and `result.firstName` only exists on `SuccessResponse`. 

If you go:

```typescript
expect(sierraUser.result).toBe(ResponseStatus.Success)
expect(sierraUser.result.firstName).toBe(firstName)
```

then the type checker complains that it doesn't know if `result.firstName` is defined.

I've switched it up a bit so we have type casting rather than documentary errors; which is marginally less bad.

jest-extended looks interesting but I can't find any examples where we've actually used it; might be one to revisit later.

jamieparkinson: here's an example - it mutates the global `expect` so a bit hard to find
https://github.com/wellcomecollection/identity/blob/main/packages/shared/sierra-client/tests/http-sierra-client.test.ts#L141

PR #361, opened 2022-07-21
Update the fields set on a new patron in Sierra

alexwlchan:


melanierogan (approved):
Excellent to have those comments/notes for context and reminders

PR #363, opened 2022-07-22
Tidy up the code for extracting names from Sierra records; add proper tests

alexwlchan:
This is a no-op refactoring change for https://github.com/wellcomecollection/wellcomecollection.org/issues/8207

* The `patron.ts` file in the shared Sierra client was a bit big and unwieldy, and doing a bunch of things.
* I've broken it up, and created a new `marc` namespace which will contain the MARC-related logic. Right now that means:

    1. the MARC field types
    2. the function `getPatronName(varFields: VarField[])`

* I've added a bunch of tests for `getPatronName()`, based on the structure of names in the real patron data. This gives us more of a safety net when working on this function, and allowed me to simplify in a couple of places where we're handling scenarios that never occur in practice. I'll highlight these inline.

* The next step will be a new function `createPatronName(name: Name): VarField`, which we can use whenever we want to set somebody's name – but that'll be a separate patch.


packages/shared/sierra-client/src/patron.ts
```diff
@@ -16,102 +22,6 @@ export function toPatronRecord(response: PatronResponse): PatronRecord {
...
-  lastName: string;
-} {
-  if (!content.trim()) {
-    return {
-      firstName: '',
-      lastName: '',
-    };
-  }
-
-  content = content.replace('100', '').replace('a|', '').replace('_', '');
```
alexwlchan, line 95: I binned this line.

The first two replacements are, AFAICT, pointless – I couldn't find any patron records where they'd apply.

The last one is bordering on unhelpful – there are patron records where an underscore has been used and looks significant, e.g. `Bonham_Carter, Helena` – it looks as if an underscore has been used in lieu of other punctuation and/or a special character. There's no explanation of why we remove it, so I decided to put them back in.

melanierogan (approved):
This is great Alex, much tidier. I also appreciated the test first and last names.

PR #364, opened 2022-07-22
Add a helper for creating the varField for a name

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8207; follows #363

Previously we had two slightly divergent implementations for creating a name varfield:

* When we create a user, we'd set field tag `n`, MARC tag 100 and subfields $a and $b
* When we updated a user, we'd set field tag `n` and subfields $a and $b, but no MARC tag 100

I think the omission of MARC tag 100 is causing us issues – e.g. in the Sierra GUI, a name without it is treated as opaque text rather than a structured MARC field.

This patch introduces a single implementation `createNameVarField()` which is used in both places, and gives us consistent behaviour.


melanierogan (approved):
✨

PR #365, opened 2022-08-04
Add an endpoint for resending verification emails

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8179

I think this will allow the identity web app to make a POST to the identity API and get a new verification email to be sent. There's a separate piece of work to actually call it from the identity web app.


jamieparkinson (commented):
This is good so far but needs the following:

- Tests
- A permission/scope to be represented in the [resource server](https://github.com/wellcomecollection/identity/blob/main/infra/scoped/auth0-resource-server.tf)
- An entry [in the authorizer](https://github.com/wellcomecollection/identity/blob/main/packages/apps/api-authorizer/src/handler.ts#L26-L47)

packages/apps/api/test/routes/users_user_id_resend_verification_email.test.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('/users/{userId}/validate', () => {
+  describe('POST /users/{userId}/resend_verification', () => {
+    it('resends a verification email for an existing user', async () => {
+      const testUser = randomExistingUser({ password: randomAlphanumeric(10) });
+      const { api } = mockedApi([testUser]);
+      const response = await withCallerId(testUser.userId)(
+        api.post(`/users/${testUser.userId}/resend_verification`)
+      );
+
+      expect(response.statusCode).toBe(204);
```
jamieparkinson, line 13: See comment on mock class - you're testing the endpoint behaviour here, but not yet whether it's doing what it says it does. You can do something along the lines of...

```typescript
const { api, clients } = mockedApi([testUser]);
const response = await withCallerId(testUser.userId)(
  api.post(`/users/${testUser.userId}/resend_verification`)
);

expect(response.statusCode).toBe(204);
expect(clients.auth0.resendVerificationEmail).toHaveBeenLastCalledWith(testUser.userId);
```

packages/shared/auth0-client/src/MockAuth0Client.ts
```diff
@@ -127,8 +127,14 @@ export default class MockAuth0Client implements Auth0Client {
...
+  private resentVerificationEmails: number[] = [];
+
   resendVerificationEmail = jest.fn(async (userId: number) => {
-    return successResponse(undefined);
+    if (this.users.has(userId.toString())) {
+      this.resentVerificationEmails.push(userId);
+      return successResponse(undefined);
+    }
+    return errorResponse('Not found', ResponseStatus.NotFound);
   });
```
jamieparkinson, line 138: What's the purpose of `resentVerificationEmails` here, as it's not being read? When you write a test for the `HttpAuth0Client` method, then you can "trust" this implementation and your test above can use the `jest.fn` to make sure it was called correctly.

alexwlchan: Ah, I think this was me trying to do what you describe in your other comment – I was planning to inspect the value and check it had been populated correctly to see that the method had been called.

alexwlchan: I've removed this.

packages/apps/api/test/routes/users_user_id_resend_verification_email.test.ts
alexwlchan: This is much nicer, added.

alexwlchan:
@jamieparkinson I think this is ready for another look.

jamieparkinson (commented):
This is good, but just to be perfectionist can you [add a test](https://github.com/wellcomecollection/identity/blob/main/packages/shared/auth0-client/tests/auth0-client.test.ts) for the new method you've added to the client (I'm aware that there are not currently tests for every method, but there should be 😬)

alexwlchan:
Test added.

I've renamed it to "send verification" email rather than "resend", because in some cases we've never sent an email in the first place.

I've also removed the scope; it turns out this is covered by the existing `update:users` scope.

alexwlchan:
Ah, sorry, didn't realise Git was waiting for me to force push. 🤦‍♀️ 

This actually has the updated code now.

packages/apps/api-authorizer/src/handler.ts
```diff
@@ -34,6 +34,9 @@ const validateRequest = resourceAuthorizationValidator({
   '/users/{userId}/deletion-request': {
     PUT: allOf(isSelf, hasScopes('delete:patron')),
   },
+  '/users/{userId}/send_verification': {
+    POST: isSelf,
```
jamieparkinson, line 38: Why no scope check on this?

jamieparkinson (approved):

packages/apps/api-authorizer/src/handler.ts
alexwlchan: Because I forgot. 🤦‍♀️

PR #370, opened 2022-08-15
Throw a generic error for all error cases 

davidpmccormick:
So that we're not telling users about the specifics of our infrastructure.

### DIFF
--- a/packages/apps/auth0-database-scripts/src/login.ts
+++ b/packages/apps/auth0-database-scripts/src/login.ts
@@ -10,13 +10,8 @@ declare const configuration: {
   CLIENT_SECRET: string;
 };
 
-const notFoundPatronErrorMessage = 'We have not found the patron in Sierra';
-const otherFindingSierraPatronErrorMessage =
-  'There was some other error in finding the patron in Sierra';
-const sierraPatronValidationErrorMessage =
-  'We had an issue in running validation of the patron in Sierra';
-const markingPatronVerifiedErrorMessage =
-  'We encountered an error in marking the patron as verified in Sierra';
+const contactLibraryErrorMessage =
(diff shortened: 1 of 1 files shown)


alexwlchan (commented):
1. Let's add a comment explaining why this is generic.
2. Maybe tell the users "Please check your email address and password are correct"? Since that seems to be today's failure mode.

packages/apps/auth0-database-scripts/src/login.ts
```diff
@@ -10,13 +10,8 @@ declare const configuration: {
...
 
-const notFoundPatronErrorMessage = 'We have not found the patron in Sierra';
-const otherFindingSierraPatronErrorMessage =
-  'There was some other error in finding the patron in Sierra';
-const sierraPatronValidationErrorMessage =
-  'We had an issue in running validation of the patron in Sierra';
-const markingPatronVerifiedErrorMessage =
-  'We encountered an error in marking the patron as verified in Sierra';
+const contactLibraryErrorMessage =
+  'There was an error. Please email us at library@wellcomecollection.org if this problem persists.';
```
jamieparkinson, line 14: This isn't quite right - the right security precaution is just not to distinguish between incorrect username or password, or missing users. We definitely should be telling people when they have entered the wrong credentials! The old message seemed fine to me:
```
We don't recognise the email and/or password you entered. Please check your entry and try again.
```

I'm not sure if the plain `Error` messages get exposed - I think it might just be the `WrongUsernameOrPasswordError` that supports that - but the "please email us" _would_ be appropriate for those, so let's do that.

jamieparkinson (changes requested):

jamieparkinson (approved):

PR #371, opened 2022-08-15
Remove some of the 'account management system' naming

alexwlchan:
There is no service called "Account Management System"; there's the identity web app in the dotorg repo. This name is a leftover Digirati-ism.

It'll be a bit of a faff to remove it completely, but we can start moving in that direction for resources that support easy renaming (including applications in Auth0 and API Gateway API keys).

I was actually trying to create a documented, turnkey process for devs to create clients that they could use when running the identity web app locally, but I got distracted by bad naming.

~This is currently applied in staging only; if we like it I'll apply it in prod then delete all the `moved` blocks.~ I had to apply in prod to deploy the "resend email verification" stuff, and this seemed harmless enough.


jamieparkinson (approved):
👏

PR #378, opened 2022-08-26
Enable more specific errors for 'User already exists'

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8202

This is the new error when you try to sign up with an already-used email address:

Screenshot 2022-08-26 at 08 28 25


alexwlchan:
Should have said previously: this is already applied in stage; applying in prod before I merge.

PR #381, opened 2022-10-04
Improve the error message when somebody has multiple patron records

alexwlchan:
Last night we saw a smattering of errors when somebody tried to sign up. The underlying issue wasn't that they already had a Sierra account (which we already handle correctly); it's that they had **two**. This should be impossible – we cleaned up all the duplicate accounts before launching the new user sign-in, and Auth0 should prevent this from happening – but it did, and we give a less than helpful error:

Screenshot 2022-10-04 at 08 23 11

This patch adds a `DuplicateUsers` response status from the `getRecordByEmail()` method on the Sierra client, and then we can respond to it with a more actionable error message:

Screenshot 2022-10-04 at 08 22 12

This copy is based on a similar error which already works correctly on the sign-in screen:

Screenshot 2022-10-04 at 08 22 03

I don't know why one error is in a red box and one is in inline text and I'm not going to spend the time to work it out; this should help the handful of users who are affected by this issue – and possibly prompt LE&E to work out where this is coming from.

### DIFF
--- a/packages/apps/auth0-database-scripts/src/create.ts
+++ b/packages/apps/auth0-database-scripts/src/create.ts
@@ -34,6 +34,11 @@ async function create(user: Auth0UserWithPassword) {
     user.email,
     user.password
   );
+
+  console.log(
+    `Received create patron response = ${JSON.stringify(createPatronResponse)}`
+  );
+
   if (createPatronResponse.status === ResponseStatus.UserAlreadyExists) {
     console.log('CREATE PATRON IN SIERRA ERRORS - USER ALREADY EXISTS');
     throw new ValidationError(user.email, userAlreadyExistsMessage);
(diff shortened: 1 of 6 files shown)


packages/apps/auth0-database-scripts/src/create.ts
```diff
@@ -34,6 +34,11 @@ async function create(user: Auth0UserWithPassword) {
     user.email,
     user.password
   );
+
+  console.log(
+    `Received create patron response = ${JSON.stringify(createPatronResponse)}`
+  );
```
alexwlchan, line 40: This is because we replace the original response when we throw our `'Please use a more complex password'` error, and it'd be useful to log the underlying Sierra response. It appears in the Auth0 log console.

packages/apps/auth0-database-scripts/src/get_user.ts
```diff
@@ -24,6 +25,11 @@ async function getUser(email: string): Promise<Auth0User | undefined> {
     return patronRecordToUser(patronRecord);
   } else if (patronRecordResponse.status === ResponseStatus.NotFound) {
     return undefined;
+  } else if (patronRecordResponse.status === ResponseStatus.DuplicateUsers) {
+    throw new ValidationError(
```
alexwlchan, line 29: If I change this to an `Error` I get the generic "Unknown error occurred" red box from Auth0. 🤷‍♀️

PR #388, opened 2023-01-13
Use the new Lambda module

jamieparkinson:
Applied in stage; needs a little care to apply without downtime because Lambda versions are liable to refer to old (ie deleted) IAM roles.


infra/scoped/lambda.tf
```diff
@@ -1,10 +1,12 @@
...
-  handler       = "index.lambdaHandler"
-  role          = aws_iam_role.identity_api_gateway_lambda_role.arn
-  runtime       = "nodejs12.x"
+module "authorizer_lambda" {
+  source = "git@github.com:wellcomecollection/terraform-aws-lambda.git?ref=v1.1.1"
+
+  name = "identity-authorizer-${terraform.workspace}"
+
+  handler = "index.lambdaHandler"
+  runtime = "nodejs12.x"
```
alexwlchan, line 9: Not for this PR, but do we have a ticket to get these on a newer version of Node? I think AWS is phasing out Node 12 soon.

alexwlchan (approved):

infra/scoped/lambda.tf
jamieparkinson: Good point - no, and actually it's being phased out pretty soon. I'll look at it now as it should be quite straightforward

PR #389, opened 2023-01-19
Upgrade to Node 18

jamieparkinson:
Well - this wasn't as simple as I thought it would be...
- Everything works fine with Node 18
- _but_ - there was an error getting swallowed in the tests which got exposed by the upgrade, leading me to realise that the registration hooks weren't behaving as I thought they were. Something to look into afterwards, nothing here changes behaviour (just exposes it).
- Fedora had some weird issue with Node 18 segfaulting on startup; I am not really familiar with CentOS and am much happier with Debian so I switched out the build environment image
- This exposed some oddities with the way the docker-compose buildkite plugin uses shells, which took a while to fix


alexwlchan (approved):
Ancient programmer curse: "may you think a task will be easy"

Issue #400, opened 2024-02-06
More granularity in user roles from patron type


kenoir:
In order to better distinguish user roles when offering access to restricted images we should:

Update our mapping from patron type to role to be more granular

https://github.com/wellcomecollection/identity/blob/main/packages/shared/sierra-client/src/patron.ts#L29, 

We think the user data will get updated when someone logs in the next time, we should confirm this is the case. If so we can decide how to map the patron type to role our side and communicate the appropriate role to role mapping in the style of the example in this RFC: https://github.com/dlcs/protagonist/blob/main/docs/rfcs/008-more-access-control-oidc-oauth.md#role-provider---oidc

See: https://wellcome.slack.com/archives/CBT40CMKQ/p1707148320736249

jcateswellcome:
Just moved this to the 'digital collections' board, as I think we will discuss in our planning session. I'd like to have this resolved this quarter, I think, as it relates to work in the previous quarter and the Digirati work package.

kenoir:
> I believe this will also need a change to the OpenAthens config so that staff can continue to log in using wc.org as an identity provider

@jamieparkinson can you expand on this? I'm unsure on how a new Role here would impact the OpenAthens login?

jcateswellcome:
Collections information have confirmed they are happy with the proposed approach of adding a new patron type. I'll confirm the next steps with product lines for doing that, then we can update the mapping.

jamieparkinson:
Missed the q above - my memory is that journal access for Wellcome staff can be via AD, but we also configured OpenAthens to map the Staff patron type so that they can log in via wc.org. That mapping also makes it so that the self-registered patron type can't access journals at all.

jcateswellcome:
Is part of restricted items access for staff [epic](https://github.com/wellcomecollection/identity/issues/402)

PR #403, opened 2024-06-06
Add new StaffWithRestricted patron type

kenoir:
## What does this change?

For https://github.com/wellcomecollection/identity/issues/400

In order to better restrict who has access to certain images, this change adds a new patron type for staff that should have access to restricted images (reducing this from all staff to only staff with this role).

This change will require an update to the IIIF API mappings here: https://github.com/wellcomecollection/iiif-builder/blob/main/src/AuthTest/AuthTest/Controllers/HomeController.cs#L38, requested here; https://wellcome.slack.com/archives/CBT40CMKQ/p1717676657432939. The new mapping should allow only `StaffWithRestricted` access to restricted images.

In addition this will require an update to the OpenAthens configuration that currently uses the `Staff` role from Auth0 to distinguish who should have access to certain journals. New configuration will need to include also the  `StaffWithRestricted` type.

## How to test

- [ ] Run the tests, do they pass?
- [ ] Deploy this change, ask a test user to login - is their patron type updated in Auth0?
- [ ] When the IIIF API changes have occurred, can only the new staff type get access to restricted images?

## How can we measure success?

Only staff users who need access to restricted images should have access to those images.

## Have we considered potential risks?

We are changing authentication for some users, but this should reduce permissions for staff rather than increase it. We should see if we can test this change on stage before releasing.

### DIFF
--- a/packages/shared/sierra-client/src/patron.ts
+++ b/packages/shared/sierra-client/src/patron.ts
@@ -22,7 +22,12 @@ export function toPatronRecord(response: PatronResponse): PatronRecord {
   };
 }
 
-export type Role = 'Reader' | 'Staff' | 'SelfRegistered' | 'Excluded';
+export type Role =
+  | 'Reader'
+  | 'Staff'
+  | 'SelfRegistered'
+  | 'Excluded'
+  | 'StaffWithRestricted';
 
 // You can find an up-to-date list of patronType codes and descriptions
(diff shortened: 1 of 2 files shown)


paul-butcher (approved):

kenoir:
Closes:  https://github.com/wellcomecollection/identity/issues/400

PR #412, opened 2025-02-05
Update terraform version

kenoir:
## What does this change?

The version of terraform we were using locally to test changes has caused a state conflict (see [this build error](https://buildkite.com/wellcomecollection/identity-deployment/builds/1421#0194d1e3-f429-4ecf-8b9f-a8c56473f750/279-377)).

Updating the version of terraform used in CI to a more recent version, and associated scripts that [override backend config](https://github.com/wellcomecollection/identity/compare/rk/update-tf?expand=1#diff-c9068e498ec07a078824cccaa510acebbfdea618c4cd96a6aa9f08508d379f95R21) now at a new path.

Updating the buildkite build image requires building a new image locally and pushing that to ECR, which has been done in this case, overwriting the previous tag (the sha of the last current working image is sha256:33b6f30414202429e9c9121c6bb4e8190aa62fbee9972cfa9c30d12795ef36f2 if we need to go back).

While updating the image I ran into the issue that the node 18 image has changed and won't allow installation of libs with pip without the `--break-system-packages` flag, and that `lint-staged` needs a git config change. 

_I pushed a version over the top, hoping we could fix it in place, we can't!_

## How to test

Will this deploy to stage after being merged?

## How can we measure success?

Identity deployments can proceed.

## Have we considered potential risks?

We should not deploy changes to prod until this is resolved (but we can't because the deploy is broken).

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,7 +1,7 @@
 steps:
   - name: "Run linting"
     command:
-      - ". /app/.buildkite/scripts/init-environment.sh"
+      - "source /app/.buildkite/scripts/init-environment.sh"
       - "/app/.buildkite/scripts/linting.sh"
     plugins:
       - wellcomecollection/aws-assume-role#v0.2.2:
@@ -14,7 +14,7 @@ steps:
 
   - name: "Run tests"
     command:
-      - ". /app/.buildkite/scripts/init-environment.sh"
+      - "source /app/.buildkite/scripts/init-environment.sh"
       - "yarn install --frozen-lockfile"
       - "yarn build"
       - "yarn test"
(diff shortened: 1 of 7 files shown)


infra/modules/ecs_service/main.tf
```diff
@@ -1,25 +1,25 @@
 module "log_router_container" {
-  source    = "git::github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/firelens?ref=v3.13.1"
+  source    = "git::https://github.com/wellcomecollection/terraform-aws-ecs-service.git//modules/firelens?ref=v3.13.1"
```
kenoir, line 2: Get modules via HTTPs to avoid authentication problem with SSH in CI, these repos are public in any case.

.buildkite/pipeline.yml
```diff
@@ -1,6 +1,7 @@
 steps:
   - name: "Run linting"
     command:
+      - ". /app/.buildkite/scripts/init-environment.sh"
```
agnesgaroux, line 4: Do we need to init the env for the linting job? Not a blocker!

agnesgaroux (approved):

.buildkite/pipeline.yml
agnesgaroux: We do 👍

